PR #2, opened 2017-11-28
Use $Shape<T> instead of optional fields

MichaelDeBoey:
Ref: https://github.com/facebook/flow/pull/5251/files#diff-453a8ac419736c97694ea5b93d1965e0R120

### DIFF
--- a/src/types.js.flow
+++ b/src/types.js.flow
@@ -1,40 +1,40 @@
 // @flow
-export type Subscription = { [string]: boolean }
-export type Subscriber<V> = (value: V) => void
+export type Subscription = { [string]: boolean };
+export type Subscriber<V> = (value: V) => void;
 
-export type FormSubscription = {
-  active?: boolean,
-  dirty?: boolean,
-  error?: boolean,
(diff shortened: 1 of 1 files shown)


erikras:
Fascinating! I had no idea that was an option. That's pretty bleeding edge to reference a docs PR! 😎 

Thanks!

MichaelDeBoey:
@erikras There's no script to run `prettier`, or did I overlooked it? 😕

erikras:
It's _supposed_ to run on `precommit`, but sometimes it doesn't. The setup looks identical on `react-final-form`, and it works on one and not on the other... 🤷‍♂️

MichaelDeBoey:
@erikras `$Shape` already exists in `flow`, but just isn't documented yet... 😕
@aikar's PR (facebook/flow#5251) just adds a good explanation, so was a good reference for this PR 🙂

erikras:
Published in [`v1.1.0`](https://github.com/erikras/final-form/releases/tag/v1.1.0).

PR #16, opened 2017-12-04
Fixed bug in sync error blocking submissions

erikras:
Fixes https://github.com/erikras/react-final-form/issues/17.


erikras:
Published in [`v1.2.1`](https://github.com/final-form/final-form/releases/tag/v1.2.1).

PR #24, opened 2017-12-06
Added Decorator type

erikras:
Just a flow type, to enable things to come...


erikras:
Published in [`v1.3.0`](https://github.com/final-form/final-form/releases/tag/v1.3.0).

PR #28, opened 2017-12-10
Added data prop to published field state

erikras:
Wasn't actually copying `data` value from private to public field state.


erikras:
Published in [`v1.3.1`](https://github.com/final-form/final-form/releases/tag/v1.3.1).

PR #29, opened 2017-12-10
Added check that a field is registered before setting error

erikras:
Fixes https://github.com/final-form/react-final-form/issues/49.


erikras:
Published in [`v1.3.1`](https://github.com/final-form/final-form/releases/tag/v1.3.1).

PR #30, opened 2017-12-12
Add TypeScript typings

mark-raymond:
I manually translated the Flow typings to TypeScript typings.

The conversion was largely obvious, apart from `FormApi.submit` and `Config.onSubmit`, where I'm not sure whether all the `undefined`s are wanted; I think they're a correct literal translation, but might not be the intention.

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -55,7 +55,7 @@ module.exports = {
       description: 'Generates table of contents in README',
       script: 'doctoc README.md'
     },
-    copyTypes: npsUtils.copy('src/*.js.flow dist'),
+    copyTypes: npsUtils.copy('src/*.js.flow src/*.d.ts dist'),
     lint: {
       description: 'lint the entire project',
       script: 'eslint .'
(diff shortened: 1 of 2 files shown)


erikras:
Nice! The `undefined`s are because `onSubmit` doesn't _have_ to return anything.

Great work.

erikras:
@mark-raymond Seems like this also needs some equivalent of [this line](https://github.com/final-form/final-form/blob/master/package-scripts.js#L58), to copy the `.ts` files to `dist`, no?

mark-raymond:
Good point, I've updated that line 😄

erikras:
Can you do `react-final-form`, too? 🙏

mark-raymond:
I'll have a go! I'll need to wait until `final-form` is released with this in, so that `react-final-form` can reference the `final-form` typings.

erikras:
Published in [`v1.3.2`](https://github.com/final-form/final-form/releases/tag/v1.3.2).

PR #31, opened 2017-12-13
Fixed deep array field-level validation bug

erikras:
Fixes https://github.com/final-form/react-final-form-arrays/issues/3.


erikras:
Published in [`v1.3.3`](https://github.com/final-form/final-form/releases/tag/v1.3.3)

PR #33, opened 2017-12-14
Fixed array deletion bug

### LINKED ISSUE #32: Removes object from array when all its fields are empty
See sandbox from https://github.com/final-form/react-final-form-arrays/issues/3#issuecomment-351415186. Add a few customers, and then type into the "last name" field and then remove the text from that field. The whole customer vanishes. 🤦‍♂️

erikras:
Fixes #32.


erikras:
Published in [`v1.3.5`](https://github.com/final-form/final-form/releases/tag/v1.3.5).

PR #43, opened 2017-12-17
Update docs' TOC

MichaelDeBoey:


erikras:
Thanks for these, but I changed my mind. #45.

PR #44, opened 2017-12-17
Update TypeScript FormApi type

MichaelDeBoey:


erikras:
Thanks for these, but I changed my mind. #45.

PR #46, opened 2017-12-18
Custom isEqual and validateFields field config

### LINKED ISSUE #37: Validate only changed field
Feature request

Hey :) First thanks for this great library, it is a real pleasure working with it, but let me go to the point. I have a large form containing ~20 fields. Each field has its own validation function passed in `form.registerField`, so current behaviour is as follows. If I change the value of any field, it will trigger validation of every registered field, which (in my opinion) and in most of my cases is unnecessary, especially when some field may use async validation (via ajax). I would like to kindly ask you to consider changing this behaviour. Perhaps a new option could be added to `registerField` e.g:

```js
  form.registerField(
  'someField',
  (fieldState) => {
    this.someField = fieldState
  },
  { value: true, valid: true },
  (value) => !!val,
  validateFields?: boolean | Array<string> 
  // if it's false will only trigger validation of that field, 
  // if true will trigger validation of every field, 
  // or if it's an array will validate specified fields 
)
```

When field value is updated, only trigger validation function of the updated field

Thank you for looking into this.

erikras:
Fixes #37 and https://github.com/final-form/react-final-form/issues/59.

⚠️ BREAKING CHANGE ⚠️

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ form.submit() // only submits if all validation passes
 
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
 * [Examples](#examples)
   * [Simple React Example](#simple-react-example)
 * [Libraries](#libraries)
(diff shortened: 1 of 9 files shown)


erikras:
Published in [`v2.0.0`](https://github.com/final-form/final-form/releases/tag/v2.0.0).

PR #53, opened 2017-12-29
Update FormState TS definition

circlingthesun:
Defining the 'values' property of FormState as 'object' forces one to reference keys via bracket notation. Using dot notation results in: Property 'myprop' does not exist on type 'object'. Redefining 'values' as '{ [key: string]: any }' lets one use dot notation.


erikras:
Published in [`v3.0.1`](https://github.com/final-form/final-form/releases/tag/v3.0.1).

PR #54, opened 2018-01-08
[typescript] Use interfaces where possible for reuse cases

rosskevin:
When a developer wants to reuse types, one cannot pick from or extend any `type`, but can from any `interface`.

This PR is neutral in typing results but provides a better case for reuse/extension of types in a developer's application.

I'll be submitting a similar PR for react-final-form shortly.


erikras:
Published in [`v3.0.1`](https://github.com/final-form/final-form/releases/tag/v3.0.1).

PR #57, opened 2018-01-11
Added dirtySinceLastSubmit flag for fields and form

### LINKED ISSUE #4: Form Submission Error Display
Hi, thanks so much for this library! Looks really nice and minimal.

I'm designing an interaction in which form errors are shown to the user after they attempt to submit a form.

If there are errors, I want the errors to stay on the screen until the next time the user attempts the submission. In other words, I do not want an error to appear/disappear as a user types the text that swaps it from invalid to valid.

This is not currently possible with Final Form as I understand it. If a form field is `invalid` at the point of one submission,  making it `valid` will wipe away its former `invalid`ness, removing the error display. There is no way to model "this field was previously submitted, and was invalid at that time" for display purposes. `submitFailed` is not `true`, because the submit action never happened due to the presence of errors.

Is there a way to solve this problem with Final Form today?

Thanks again!

### LINKED ISSUE #38: Blur and focus actions
**bug report** or **feature request** - hard to tell ;)

When I use `blur` and `focus` actions, they affect only `active` state of a given field. It would be great, if it could be used also to REALLY focus or blur fields, which sometimes is nice to do programmatically. For instance, lets say you have change password form, and a user sets `password` and `repeatPassword` fields, and clicks enter which submits the form. On submission success, I reset the form. The problem? Repeat password is still focused, and when it is blurred by any user action, it will trigger `this field is required` error message. Here, `blur('repeatPassword')` callled just before `reset` would fix this, but unfortunately it doesn't work for now.

### LINKED ISSUE #48: Cannot submit form after returning an error from onSubmit
Bug

Hey :) please take a look at this example: https://codesandbox.io/s/mq3opx9vo9. 
After submitting a form I return some error from `onSubmit` handler and display it on the page. The problem is that after setting form error(s), form becomes invalid and I cannot send form again. Even calling `form.reset()` (although I do not want to reset fields values to their initial ones)  doesn't help. 

Don't mark form as invalid when settings submitError/submitErrors

erikras:
Fixes #4.
Fixes #38.
Fixes #48.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,8 +407,11 @@ A place for arbitrary values to be placed by mutators.
 
 #### `dirty?: boolean`
 
-`true` when the value of the field is `!==` the initial value, `false` if the
-values are `===`.
+`true` when the value of the field is not equal to the initial value (using the [`isEqual`](#isequal-a-any-b-any--boolean) comparator provided at field registration), `false` if the values are equal.
+
+#### `dirtySinceLastSubmit?: boolean`
+
+`true` when the value of the field is not equal to the value last submitted (using the [`isEqual`](#isequal-a-any-b-any--boolean) comparator provided at field registration), `false` if the values are equal.
 
 #### `error?: any`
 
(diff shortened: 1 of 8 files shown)


erikras:
Published in [`v3.1.0`](https://github.com/final-form/final-form/releases/tag/v3.1.0).

PR #59, opened 2018-01-12
Remove unnecessary `flow` dev dependency

pepakriz:


erikras:
Why is this unnecessary? Because `flow-bin` contains it?

pepakriz:
Because `flow` is different library with similar name. See https://www.npmjs.com/package/flow vs https://www.npmjs.com/package/flow-bin

erikras:
Duh. 🙄 Must have been an `npm install` typo some time ago... Thanks!

pepakriz:
Yeah, initial commit contains this typo :smile:

PR #60, opened 2018-01-12
Fix ts defs and add index.d.test.ts

brikou:
Based on issue https://github.com/final-form/react-final-form/issues/108 here is a fix for ts regression described on screenshot of this issue.

I've added a demo ts file so you can run compilation on CI and prevent any other regression (or at least if tested). For now the test cover only form creation and I havn't added the required command in build/ci script.

the command is just `node_modules/.bin/tsc`

I've reverted most changes from #54 because it causes regression and this is not the way to use typescript. If you want a partial interface then just use `Partial<MyInterface>`, if you want only a single def then you use `MyInterface['singleField']` adding `?` is not the way to do.

Tell me if you have question.

Thanks :+1:

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,218 +1,216 @@
-declare module 'final-form' {
-  export interface Subscription {
-    [key: string]: boolean
-  }
-  export type Subscriber<V> = (value: V) => void
-  export type IsEqual = (a: any, b: any) => boolean
-
-  export interface FormSubscription extends Partial<Subscription> {
-    active: boolean
-    dirty: boolean
-    dirtySinceLastSubmit: boolean
-    error: boolean
-    errors: boolean
-    initialValues: boolean
-    invalid: boolean
-    pristine: boolean
-    submitError: boolean
-    submitErrors: boolean
-    submitFailed: boolean
-    submitSucceeded: boolean
-    submitting: boolean
-    valid: boolean
-    validating: boolean
-    values: boolean
-  }
-
-  export interface FormState {
-    active: string
-    dirty: boolean
-    dirtySinceLastSubmit: boolean
-    error: any
-    errors: object
-    initialValues: object
-    invalid: boolean
-    pristine: boolean
-    submitError: any
-    submitErrors: object
-    submitFailed: boolean
-    submitSucceeded: boolean
-    submitting: boolean
-    valid: boolean
-    validating: boolean
-    values: { [key: string]: any }
-  }
-
-  export type FormSubscriber = Subscriber<FormState>
-
-  export interface FieldState {
-    active?: boolean
-    blur: () => void
-    change: (value: any) => void
-    data?: object
-    dirty?: boolean
-    dirtySinceLastSubmit?: boolean
-    error?: any
-    focus: () => void
-    initial?: any
-    invalid?: boolean
-    length?: number
-    name: string
-    pristine?: boolean
-    submitError?: any
-    submitFailed?: boolean
-    submitSucceeded?: boolean
-    touched?: boolean
-    valid?: boolean
-    value?: any
-    visited?: boolean
-  }
-
-  export interface FieldSubscription extends Partial<Subscription> {
-    active: boolean
-    data: boolean
-    dirty: boolean
-    dirtySinceLastSubmit: boolean
-    error: boolean
-    initial: boolean
-    invalid: boolean
-    length: boolean
-    pristine: boolean
-    submitError: boolean
-    submitFailed: boolean
-    submitSucceeded: boolean
-    touched: boolean
-    valid: boolean
-    value: boolean
-    visited: boolean
-  }
-
-  export type FieldSubscriber = Subscriber<FieldState>
-
-  export type Unsubscribe = () => void
-
-  type FieldValidator = (value: any, allValues: object) => any | Promise<any>
-  type GetFieldValidator = () => FieldValidator
-
-  export interface FieldConfig {
-    isEqual?: IsEqual
-    getValidator?: GetFieldValidator
-    validateFields?: string[]
-  }
-
-  export type RegisterField = (
-    name: string,
-    subscriber: FieldSubscriber,
-    subscription: FieldSubscription,
-    config: FieldConfig
-  ) => Unsubscribe
-
-  export interface InternalFieldState {
-    active: boolean
-    blur: () => void
-    change: (value: any) => void
-    data: object
-    error?: any
-    focus: () => void
-    isEqual: IsEqual
-    lastFieldState?: FieldState
-    length?: any
-    name: string
-    submitError?: any
-    pristine: boolean
-    touched: boolean
-    validateFields?: string[]
-    validators: {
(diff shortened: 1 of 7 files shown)


erikras:
cc @rosskevin 

Generally looks good. See if you can figure out how to add it to `package-scripts.js`, alongside `lint` so that it is run on every `npm start validate` call.

brikou:
@erikras done :)

rosskevin:
@brikou can you clarify your comment?

> I've reverted most changes from #54 because it causes regression and this is not the way to use typescript. If you want a partial interface then just use Partial<MyInterface>, if you want only a single def then you use MyInterface['singleField'] adding ? is not the way to do.

I converted:

```diff
- export type FormSubscription = Partial<{
 -  active: boolean
 -  dirty: boolean
 -  error: boolean
 -  errors: boolean
 -  initialValues: boolean
 -  invalid: boolean
 -  pristine: boolean
 -  submitError: boolean
 -  submitErrors: boolean
 -  submitFailed: boolean
 -  submitSucceeded: boolean
 -  submitting: boolean
 -  valid: boolean
 -  validating: boolean
 -  values: boolean
 -}> &
 -  Subscription
+export interface FormSubscription extends Subscription {
 +  active?: boolean
 +  dirty?: boolean
 +  error?: boolean
 +  errors?: boolean
 +  initialValues?: boolean
 +  invalid?: boolean
 +  pristine?: boolean
 +  submitError?: boolean
 +  submitErrors?: boolean
 +  submitFailed?: boolean
 +  submitSucceeded?: boolean
 +  submitting?: boolean
 +  valid?: boolean
 +  validating?: boolean
 +  values?: boolean
 +}
 +
```

Which are functionally equal.

src/index.d.ts
```diff
@@ -1,217 +1,218 @@
...
-  submitError?: boolean
-  submitErrors?: boolean
-  submitFailed?: boolean
-  submitSucceeded?: boolean
-  submitting?: boolean
-  valid?: boolean
-  validating?: boolean
-  values?: boolean
-}
+declare module 'final-form' {
```
rosskevin: `declare module 'final-form'` is unnecessary and can cause pain when people privately publish packages e.g. scoped branch/temp release.  Please remove this unless there is a good reason to add it.

Also, this simple change makes a huge diff, so it is hard to see what the actual diff you are proposing.

rosskevin:
@brikou thanks for seeding tests!  I'll definitely add to them as I see anything pop up.

src/index.d.ts
```diff
@@ -1,217 +1,218 @@
...
+    submitFailed: boolean
+    submitSucceeded: boolean
+    submitting: boolean
+    valid: boolean
+    validating: boolean
+    values: boolean
+  }
 
-export interface FormState {
-  // all values are optional because they must be subscribed to
```
rosskevin: Your changes are not equivalent to the previous definitions, so we need to understand why.  Perhaps there is a design decision here to be made, specifically what was there before `// all values are optional because they must be subscribed to` is now no longer there.

Now all options are guaranteed to be there, which is actually the default and more accurate.  Another argument to use it as you have proposed is that otherwise we would need to check for undefined.  So I'm inclined to go with these changes but would like a proper discussion/argument to make this change - at least for the changelog.

rosskevin:
Regarding codacy's error report, @brikou how about adding `ts` files to prettier to make sure they are consistent?

src/index.d.ts
brikou: You are right, I'll remove it

brikou:
@rosskevin , thanks for your review

> Regarding codacy's error report, @brikou how about adding ts files to prettier to make sure they are consistent?

It is already done in package.json

> I converted: ... Which are functionally equal. I'm not claiming they are correct, just that the point of #54 was to convert to interfaces.

No it is not the same, as it leads to compilation error because it is extending an interface which does not allow undefined value. 

image

src/index.d.ts
```diff
@@ -4,23 +4,23 @@ export interface Subscription {
...
-  pristine?: boolean
-  submitError?: boolean
-  submitErrors?: boolean
-  submitFailed?: boolean
-  submitSucceeded?: boolean
-  submitting?: boolean
-  valid?: boolean
-  validating?: boolean
-  values?: boolean
+export interface FormSubscription extends Partial<Subscription> {
```
rosskevin: `Subscription` is an indexed object, and therefore has no required fields.  Why use `Partial` here?

src/index.d.ts
```diff
@@ -68,23 +68,23 @@ export interface FieldState {
...
-  length?: boolean
-  pristine?: boolean
-  submitError?: boolean
-  submitFailed?: boolean
-  submitSucceeded?: boolean
-  touched?: boolean
-  valid?: boolean
-  value?: boolean
-  visited?: boolean
+export interface FieldSubscription extends Partial<Subscription> {
```
rosskevin: Same question about use of `Partial` when it has now effect (that I can see)

rosskevin:
Sorry @brikou, your brevity in referring to `it` has confused me.  

```js
Partial<{foo: string}> === {foo? string}
```

does it not?

rosskevin:
@erikras - after looking at this, I'm not sure the base set of state vs subscription types have optionality correct.  I'm reminded of it because the optionality in this PR sets everything to provided in the `FormSubscription` which can be seen two ways:
1. by default all of these are subscribed and therefore provided (so this change is correct); OR
2. when you subscribe, _none_ are provided except what you chose.

But we can't see that at static analysis time, so I think the change in optionality (option 1 above) on `FormSubscription` properties by @brikou is right.  So if this is right, shouldn't we change `FormState` to match? (which I think was in @brikou's original proposal)  For example:

```diff
export interface FormState {
-  // all values are optional because they must be subscribed to
-  active?: string
-  dirty?: boolean
+  active: string
+  dirty: boolean
}
```
As I have seen so far from a user perspective, these keys seem to be provided by default.

src/index.d.ts
brikou: Don't know if this is the best fix, but you can `git apply` my diff (https://patch-diff.githubusercontent.com/raw/final-form/final-form/pull/60.diff) and try to revert this, you'll see it breaks.

src/index.d.ts
brikou: Please give a try by git applying my changes and try both (on the same time run tsc --watch and see). Maybe you can find a better fix because I'm not used to subscription that much.

rosskevin:
Thanks @brikou - I will do - I'm on a flight and can't seem to pull from git via their wifi. I'll probably get to this tomorrow.

brikou:
@rosskevin Thank you, tell me if there is anything I can do.

rosskevin:
Thanks @brikou - I expanded your tests and made some changes in #61.  Your original assertions work.

brikou:
:+1: @rosskevin

erikras:
Done in #61.

erikras:
Published #61 in [`v4.0.1`](https://github.com/final-form/final-form/releases/tag/v4.0.1).

PR #61, opened 2018-01-14
Typescript tests and fixes

rosskevin:
Supersedes #60 

Based on the conversation in #60, I have continued @brikou's work and expanded the test case.  

- I did a runtime capture of state and added these assertions in the test.
- I broke out the indexer signature as `AnyObject` for easier referencing/testing
- I changed optionality of many members as provided instead of optional, as by default they are populated.  
- I addressed my concerns about `Partial` and made sure @brikou's original tests run.

While it's likely not perfect, I think this is an improvement and more accurate representation of types by default.  Here's the screen capture of values I used for `FormState`.

default

Thanks to @brikou for getting this started.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3213,12 +3213,6 @@
         "write": "0.2.1"
       }
     },
-    "flow": {
-      "version": "0.2.3",
-      "resolved": "https://registry.npmjs.org/flow/-/flow-0.2.3.tgz",
-      "integrity": "sha1-+Npl76JJEn7Jk3aiiJZXKpeV0a8=",
-      "dev": true
-    },
     "flow-bin": {
       "version": "0.61.0",
       "resolved": "https://registry.npmjs.org/flow-bin/-/flow-bin-0.61.0.tgz",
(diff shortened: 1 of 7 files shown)


rosskevin:
@erikras - seems like codacy is not configured the same as prettier - regarding semicolons. I don't see a codacy config file so I'm not sure where you configured it.

erikras:
Looks good. This is ready to merge?

(Nevermind Codacy)

rosskevin:
This is good to merge, and you can close #60

brikou:
@erikras and @rosskevin thanks, @erikras can you release a new version bugbix release as current break my deploy (as ts compilation breaks). Thank you very much

erikras:
Published in [`v4.0.1`](https://github.com/final-form/final-form/releases/tag/v4.0.1).

rosskevin:
Ok, I just tried this via react-final-form and there are several problems in my codebase.  I'll expand the tests and perhaps add typescript tests to react-final-form as well.

erikras:
Great!

PR #64, opened 2018-01-17
Made submitFailed true when submit fails due to validation errors

### LINKED ISSUE #63: How do I tell if submit was attempted but validation failed
The behaviour I'm trying to produce is to have the "submit" button enabled, even if all the fields are invalid.  Once the user clicks submit, I want to disable the button if there are validation errors.  That way I can hide all error messages until `touched`, but if the user clicks submit, all the errors show up.

I tried `submitFailed` but that doesn't seem to be set to `true` unless the form is valid when handleSubmit is called.

erikras:
Fixes #63.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -572,6 +572,8 @@ const createForm = (config: Config): FormApi => {
         Object.keys(fields).forEach(key => {
           fields[key].touched = true
         })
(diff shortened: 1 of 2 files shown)


erikras:
Published in [`v4.0.1`](https://github.com/final-form/final-form/releases/tag/v4.0.1).

PR #68, opened 2018-01-22
Ensure `mutators` is defined to avoid bad strongly typed developer casting experienc

rosskevin:
I have peeled all layers of the onion.  This is the first domino that needs to fall.

Strongly typed users when attempting to use `mutators` in a form render will have to cast as it may be `undefined`.  This can be a painful and likely unnecessary experience.  This simple change will bring joy and sunshine to the lives of typescript developers far and wide.

https://github.com/final-form/react-final-form/pull/129 depends on this, as do PRs in the arrays projects.

@erikras please review this one first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "dist/final-form.cjs.js",
   "jsnext:main": "dist/final-form.es.js",
   "module": "dist/final-form.es.js",
-  "typings": "dist/index.d.js",
+  "typings": "dist/index.d.ts",
   "files": ["dist"],
   "scripts": {
     "start": "nps",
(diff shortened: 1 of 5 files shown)


rosskevin:
I'm linking locally to verify this in the other three projects, please hold.

rosskevin:
Good to merge.

rosskevin:
@erikras If you can release this I can bump the version on the next of 3 PRs to get them all merged/released.

erikras:
Published in [`v4.0.4`](https://github.com/final-form/final-form/releases/tag/v4.0.4).

PR #69, opened 2018-01-25
Allow setting undefined array values

erikras:
Fixes https://github.com/final-form/react-final-form/issues/133.


erikras:
Published in [`v4.1.0`](https://github.com/final-form/final-form/releases/tag/v4.1.0).

PR #70, opened 2018-01-25
Added touched and visited objects to form state

erikras:
Fixes https://github.com/final-form/react-final-form/issues/132.


erikras:
Published in [`v4.1.0`](https://github.com/final-form/final-form/releases/tag/v4.1.0).

Issue #73, opened 2018-01-29
Optimizing field validation - all fields validate on any change in form value


rosskevin:
### Are you submitting a **bug report** or a **feature request**?
hmmm - not sure - feature?

### What is the current behavior?
On change of any value in the form, EVERY field validation fires.

https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L459

### What is the expected behavior?

**When** field X value changes
**Then** the field X validator is run
**And** the record validator is run

### What's your environment?
```
yarn list v1.3.2
├─ final-form-arrays@1.0.4
├─ final-form@4.1.0
├─ react-final-form-arrays@1.0.3
└─ react-final-form@3.0.5
```

### Rationale
I type one letter into any field, and the following validators fire (some of which are a bit heavier than others):

```
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate addresses[0].address1
validate addresses[0].address2
validate addresses[0].city
validate addresses[0].stateCode
validate addresses[0].postalCode
validate addresses[1].address1
validate addresses[1].address2
validate addresses[1].city
validate addresses[1].stateCode
validate addresses[1].postalCode
```

This form feels a tad slow, and I have seemingly addressed re-renders, but have tracked down a massive number of calls to validate.

<details>
  <summary>Here is the log on-load:</summary>
```
validate firstName
validate lastName
validate firstName
validate lastName
validate email
validate firstName
validate lastName
validate email
validate mobilePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate addresses[0].address1
validate firstName
validate lastName
validate email
validate mobilePhone
validate workPhone
validate homePhone
validate addresses[0].address1
validate addresses[0].address2
validate firstName
validate lastName
validate email
[...]

PR #74, opened 2018-01-29
Added pause/resume validation functions

erikras:
Allows automatic validation to be paused and resumed. Allows fix for #73.


erikras:
Published in [`4.2.0`](https://github.com/final-form/final-form/releases/tag/v4.2.0).

PR #77, opened 2018-02-07
feature/persistentSubmitErrors added prop persistentSubmitErrors

alex-shatalov:


erikras:
Can you explain your use case and motivations here? Why is this change useful?

erikras:
Closing due to inactivity.

PR #83, opened 2018-02-21
Await async validation before submission

### LINKED ISSUE #82: Form can be submitted when async validations are running

If there's any async validation happening, onSubmit will be called regardless of their status.

onSubmit shouldn't be called until all validators have resolved. FormState.valid is true even though it might not be true. 

### Sandbox Link

https://codesandbox.io/s/lv3y8mzjm

### What's your environment?

```final-form@4.2.0```

### Other information

Failing test
```js
it('should not submit if form is still validating', () => {
  const onSubmit = jest.fn()
  const form = createForm({ onSubmit });

  const username = jest.fn()
  form.registerField('username', username, { error: true }, {
    getValidator: () => () => {
      return new Promise(resolve => resolve('Error'))
    }
  })

  form.submit()
  expect(onSubmit).not.toHaveBeenCalled()
})
```

erikras:
Fixes #82.


erikras:
Published in [`v4.2.1`](https://github.com/final-form/final-form/releases/tag/v4.2.1).

PR #87, opened 2018-03-04
Add method for setting config value after form creation

ForbesLindesay:
[see final-form/react-final-form#128]

This method lets you update any of the config values after the initial form creation. Updating mutators might be a bit slow, so would be best avoided, but all the other config options should be fine to update.  This will make it fairly simple to fix the issue in react-final-form where props can get out of date.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -666,6 +666,10 @@ const createForm = (config: Config): FormApi => {
           break
         case 'validate':
           validate = value
+          runValidation(undefined, () => {
+            notifyFieldListeners()
+            notifyFormListeners()
+          })
           break
         case 'validateOnBlur':
           validateOnBlur = value
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -22,6 +22,7 @@ import type {
   InternalFormState,
   IsEqual,
   MutableState,
+  Mutator,
   Subscriber,
   Subscription,
   Unsubscribe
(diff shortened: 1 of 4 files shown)


ForbesLindesay:
@erikras let me know what you think of this approach. I can add a test if you need me to, but I want to check you're happy with the general approach first.

erikras:
Seems okay. Would need better code coverage....

ForbesLindesay:
I've added tests that should cover most of the code

src/FinalForm.js
```diff
@@ -171,28 +172,32 @@ const createForm = (config: Config): FormApi => {
   }
 
   // bind state to mutators
+  const getMutatorApi = key => (...args) => {
+    if (!mutators) {
+      throw new Error('mutators should not be null or undefined')
```
ForbesLindesay: This is here to keep `flow` happy, but is throwing codecov off.

erikras:
Doesn't this help? https://github.com/final-form/final-form/pull/87/commits/882464bad3c680720f6f600c4c9c11552349826b

erikras:
Codacy's problems still seem legit, though.

ForbesLindesay:
882464b doesn't really change anything. There's still an implicit `else` branch that's not covered. I think it's better to have `throw new Error` there, since at least then it's clear when you're reading it that the code is probably never run.

ForbesLindesay:
The first two codacy problems were legit, the third is because you have prettier set to no-semicolons, but tslint set to recommended settings.

erikras:
Published in [`v4.3.0`](https://github.com/final-form/final-form/releases/tag/v4.3.0).

PR #90, opened 2018-03-14
Added semaphore to schedule notifications after all notifications are complete

erikras:
Fixes https://github.com/final-form/final-form-calculate/issues/8


erikras:
Published in [`v4.3.1`](https://github.com/final-form/final-form/releases/tag/v4.3.1).

PR #92, opened 2018-03-20
Added a way to set an error on an array of fields

erikras:
You can now set an error for an array of fields.

### Example
```jsx
import { ARRAY_ERROR } from 'final-form'

...

errors.items = []
errors.items[ARRAY_ERROR] = 'You need some items'
```


erikras:

PR #93, opened 2018-03-20
Added getFieldState() to form api

### LINKED ISSUE #91: Easier way to access registered fields

Feature request.

In my current project, I'm using a custom mutator to set additional field configuration into `field.data.config`, which works great. The problem is there's no easy way with the current API to access registered fields: `form.getRegisteredFields()` just returns the names, `form.getState()` returns form state and not the field objects, `form.registerField()` returns an unsubscriber and not the field object, `form.subscribe()` is form state and not field objects, etc. 

The only approach I've found is to use a mutator like `getFields`, which simply returns `state.fields`, but this seems hacky.

Field objects are abstracted away.

Being able to access fields with something like `form.getFields()`.

### What's your environment?

Final form 4.3.1
Node 8.9.1
NPM 5.6.0

### Other information

N/A

erikras:
Fixes #91.

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,9 +631,9 @@ Changes the value of the given field.
 
 Focuses (marks active) the given field.
 
-#### `getFormState: (field: string) => FieldState`
+#### `getFieldState: (field: string) => ?FieldState`
 
-Returns the state of a specific field as it was last reported to its listeners.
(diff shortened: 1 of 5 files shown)


erikras:

PR #94, opened 2018-03-21
Fixed bug where field level validation errors were getting lost

### LINKED ISSUE #75: errors lost
bug?

Existing validations are lost on form change.  

This _may_ be related to the change in #73.

screen recording 2018-01-31 at 10 59 54 am

Note that I have tried this with the async field validation example (with updated dependencies) and **it works**, so something I'm doing differs from it.  I do not yet know what I am doing differently to cause this.

validations are not lost

### Sandbox Link
Working example with latest dependencies https://codesandbox.io/s/wkvl93qyx5

### What's your environment?
```
yarn list --pattern final-form
yarn list v1.3.2
├─ final-form-arrays@1.0.4
├─ final-form@4.2.0
├─ react-final-form-arrays@1.0.3
└─ react-final-form@3.1.0
```

## TL;DR
I could not reproduce with example, but it was not doing this yesterday before #73.  If you have any guidance as to the potential source of the difference, I'd like to try to boil this down to a reproducible test case.  So far, I have had no luck.

Any thoughts  @erikras? I'll get back to this after lunch.

erikras:
Fixes #75.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -26,9 +26,7 @@ import type {
   Subscription,
   Unsubscribe
 } from './types'
-
-export const FORM_ERROR = Symbol('form-error')
-export const ARRAY_ERROR = Symbol('array-error')
+import { FORM_ERROR, ARRAY_ERROR } from './symbols'
 export const version = '4.3.1'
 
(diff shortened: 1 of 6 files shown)


erikras:

PR #97, opened 2018-03-29
Fix the API doc for DebugFunction in README.md

stefanwille:
Hi!

It seems that the documentation in REAMDE.md for DebugFunction was in the wrong place. This PR moves it and also fixes the wording - "returns" => "receives".


stefanwille:
Thanks a lot for final-form! I found the library to be very enjoyable!

erikras:
Thanks!

Issue #98, opened 2018-03-31
Accept initialize value for `reset()` 


quicksnap:
### Are you submitting a **bug report** or a **feature request**?

Feature request

### What is the current behavior?

`reset` takes no options and calls `initialize` with initial values.

### What is the expected behavior?

I've found I need to perform this operation:

```js
initialize(mySuperDefaultValues);
reset();
```

It would be great if I could instead call `reset(mySuperDefaultValues)`

In the code, we already call `initialize`: https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L665-L671

We could just pass through an optional argument which defaults to `{}`. Easy peasy PR..

erikras:
Published fix in [`v4.5.1`](https://github.com/final-form/final-form/releases/tag/v4.5.1).

PR #99, opened 2018-03-31
Accept initial values for reset.

quicksnap:
Example solution for #98 -- needs tests; do not merge.. =D

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -662,12 +662,12 @@ const createForm = (config: Config): FormApi => {
       }
     },
 
-    reset: () => {
+    reset: (initialValues = state.formState.initialValues) => {
       state.formState.submitFailed = false
       state.formState.submitSucceeded = false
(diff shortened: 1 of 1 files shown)


erikras:
I hate to implement someone's PR in my own PR, as I think that's extremely rude, but making another PR just for test and docs seemed sloppy.

Is there a way that I could've added the tests and docs directly to this PR? I don't think I have commit permissions to a fork of my repo, but sometimes Github will let me tweak things in the web interface, so it must be possible. `git` is a little like `ffmpeg`: whatever you're trying to do is certainly possible, but only if you can figure out the syntax.

quicksnap:
@erikras I don't mind at all.. especially since I made this PR in Github's site editor =P It was really just meant as a short example.

Thanks much for actually creating the feature!

erikras:
Published fix in [`v4.5.1`](https://github.com/final-form/final-form/releases/tag/v4.5.1).

PR #100, opened 2018-04-03
Allowed initial values to be passed to reset

### LINKED ISSUE #98: Accept initialize value for `reset()`

erikras:
Closes #98.
Closes #99.

Thanks for the idea, @quicksnap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,10 +681,9 @@ that should return `undefined` if the value is valid, or an error. It can
 optionally return a `Promise` that _resolves_ (not rejects) to `undefined` or an
 error.
 
-#### `reset: () => void`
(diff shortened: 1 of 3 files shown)


erikras:

PR #108, opened 2018-04-11
Added hasSubmitErrors and hasValidationErrors

### LINKED ISSUE #80: Separate validation errors and submission errors

Feature request

Currently the form/fields have two flags: `valid` and `dirtySinceLastSubmit`.

`valid` returns if the form/field have any error (validation or submission). `dirtySinceLastSubmit` returns if the form/field was touched since the last submission.

Imagine the following scenario:

- Form is initialized, validation fails so the form can't be submitted. `valid = false` so we can use that to disable the submit button, mark the field in red or do something to indicate to the user that something needs to be done.
- The user completes the form, validation passes. `valid = true` so we can enable the submit button, mark the fields in green or do something to indicate to the user that he can submit his form.
- The user submits the form.
- The values are validated again server-side for some other things and some errors occur.
- There are some submission errors, so `valid` is now false. We can disable the button again or mark fields red since there are some errors with the form.
- The user modifies some fields.
- The submission errors are not cleared on change, so the `valid` flag is still false.
- We use the `dirtySinceLastSubmit` flag to check if the user modified some things and re-enable the button or mark fields green.

And now's the problem: we can't know if the form/field is invalid because of the submission errors or if some fields don't pass the validation.

So if the user, while modifying to be able to submit again, broke some validation rules. We have no way of knowing.

We could in the form subscribe for all the errors and check if the form is passing validation or not this way, but the whole point of the subscription system is to avoid unnecessary updates, and if you validate fields on change, it means a lot of unnecessary updates.

It's not a really big problem, but it's a bit annoying to not be able to display information correctly to users.

One great way to solve this would be to separate `valid` and `invalid` to have something like `passesValidation` and `passesSubmission` so we can differentiate and we don't need to subscribe to the whole form state.

Or just to have a flag `validates` to tell if the form passes validation or not, and then it would be easy to know why the form is invalid.

I'm not sure how hard it would be to implement something like this though.

### Sandbox Link

I'll try to do a quick demo if I get the time.

### What's your environment?

final-form v4.2.0
Opera 50 on Windows 10

### LINKED ISSUE #81: Once there is a submit error, the field never becomes valid

Bug report

I have a form for entering a password. When the form is submitted, I may get an error saying "the password is incorrect" which I return from the onSubmit handler as `{password: 'the password is incorrect'}`, to indicate that the password field is the one that is incorrect.

Unfortunately, after I do this, the form never becomes "valid" again, even after the value has been edited.

I thought the error would only be there until I edited the field, at that point, I expected it to return to the valid state again.

### What's your environment?

I'm using react-define-form

erikras:
Fixes #80 
Fixes #81 
Fixes https://github.com/final-form/react-final-form/issues/89


erikras:

PR #109, opened 2018-04-11
Use strings instead of Symbols

### LINKED ISSUE #106: [question] does ARRAY_ERROR and FORM_ERROR have to be Symbols?
This is causing me some serious headache when dealing with non es6+ browsers (ie11 😭) and I'm wondering what's the use case for those? Couldnt they easily be replaced with non-symbols?

erikras:
They created more confusion and didn't really solve a real problem. Fixes #106.


erikras:

PR #111, opened 2018-04-11
Prefer simple for loop over for-of (it doesnt have to be transpiled)

Andarist:
Kinda related to #106. Transpiled for-of depends on `Symbol`

Not only this makes it symbol-independent, it makes also the overall output a little bit smaller:
before - 4419
after - 4336 💵


erikras:

PR #112, opened 2018-04-11
Use babel loose mode

Andarist:
size comparison:
before - 4419
after - 4242

Usually savings are bigger, I guess u dont use that much of es6 that needs heavier babel transforms.


erikras:

PR #113, opened 2018-04-11
Mark all as touched when submission fails with submit errors

### LINKED ISSUE #186: Bug: Cannot read property 'active' of undefined

**bug report**

I am trying to use a slider component in my form and a conditional field which is based on the field which is controlled by the slider. 

The component I'm using is this: 
https://blueprintjs.com/docs/#core/components/sliders

Showing the conditional field works just fine, but as soon as the field should disappear again, it crashes with this error message:

TypeError
Cannot read property 'active' of undefined

### Sandbox Link

Here is a sandbox with the slider:
https://codesandbox.io/s/3y197jnzq

While trying to find the cause I found that the same happens, when the value is not set by the input, but programmatically with an timeout.
I have recreated this in a sandbox (this is a really silly example, I would never actually do this, but this was the easiest way to reproduce it) 

https://codesandbox.io/s/xr4027203o

The conditionial field shows when the value is larger than 5. You can see, that manually entering a value lower than 5 works fine, but when you click the "enableTimer" button, it starts counting down from 10 and when it reaches 5 --> crash.

erikras:
Fixes #186.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -402,7 +402,6 @@ const createForm = (config: Config): FormApi => {
   }
 
   const markAllFieldsTouched = (): void => {
-    // mark all fields as touched
     Object.keys(state.fields).forEach(key => {
       state.fields[key].touched = true
     })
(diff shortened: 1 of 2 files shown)


erikras:

PR #115, opened 2018-04-12
Allowed additional values to be appended on submit

### LINKED ISSUE #85: Feature Request: Allow multiple Submit Actions / Validation Conditions
**feature request**

The standard HTML `form` submits name and value of the button that triggered the submit. This is a long known and very useful behavior.  It allows different processing and/or validation. 

i.e.: Two buttons "Create" and  "Create and Notify" where the second triggers additional email validation and notification.

```
<button type="submit" name="myAction" value="doit1">Do it 1</button>
<button type="submit" name="myAction" value="doit2">Do it 2</button>
```

There is no way to know what action triggered the submit without manually modify the form values or provide different submit handlers or input fields.

There should be a simple/default way to differentiate between triggering actions/buttons.

```
form.submit("actionName")
//or 
form.submit({additionalField:'someValue'})
//or 
form.submit.with({additionalField:'someValue'})()
```
Ror React there could be something like this.
```
<button type="submit" onClick={handleSubmit.with({additionalField:'someValue'})}>ok</button>
```

erikras:
Fixes #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ form.submit() // only submits if all validation passes
     * [`registerField: RegisterField`](#registerfield-registerfield)
     * [`reset: (initialValues: ?Object) => void`](#reset-initialvalues-object--void)
     * [`resumeValidation: () => void`](#resumevalidation---void)
-    * [`submit: () => ?Promise<?Object>`](#submit---promiseobject)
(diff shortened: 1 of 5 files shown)


kromit:
oh noes! this should happen before the validation, not after :)

PR #118, opened 2018-04-23
Performance improvements

erikras:
Improved performance by not sending notifications when validation is suspended, such as when mounting many fields.


erikras:
Published in [`v4.6.0`](https://github.com/final-form/final-form/releases/tag/v4.6.0).

perrin4869:
Hm... Actually, this PR is messing with my application logic.
I am expecting `final-form-calculate` to fire some changes, which get unexpectedly delayed and I could trace it back to this commit. Is there any chance to revert these changes, or add an option to turn them off?
Thanks!

erikras:
@perrin4869 Can you make a sandbox to demonstrate your use case?

perrin4869:
hm... won't be easy but I'll try :)
The problem in a nutshell is that am firing a `FormAPI.change` from within a react component, and I expect `final-form-calculate` to immediately fire, but it doesn't. Then I fire another `change`, and finally validation is resumed, and the `calculate` decorator finally kicks back in, undoing my second `change`.
I expect `change` -> `decorate` -> `change`, but I get `change` -> `change` -> `decorate`.
Ill setup the sandbox now anyways!

erikras:
What component lifecycle method is your `form.change()` call in?

perrin4869:
`componentDidMount` for the first change, `componentDidUpdate` for the second `change`

erikras:
Interesting. I can see how that might get tangled with the validation pausing. A sandbox would still be very helpful for me to devise a solution.

perrin4869:
https://codesandbox.io/s/vvmk4xz660

This is it.
I want the value or `bar` to be `new bar`, but it is getting `calculated bar`

perrin4869:
Also, for some reason the form is not re-rendering after updating `foo`

Issue #124, opened 2018-05-02
Pass field metadata to validator


gustavovnicius:
### What is the expected behavior?
The field `validate` signature should be:
`validate?: (value: ?any, allValues: Object, meta: Object) => ?any` since there are a few validations that could benefit from it

Examples:
While composing a `required` and a `unique` validations:

New entity form:
  - if the value is empty, it fails `required` (`{ touched: true, dirty: false }`)
  - if the value is filled and it's not unique, it fails `unique` (`{ touched: true, dirty: true }`)
  - otherwise, valid (`{ touched: true, dirty: true }`)

Now, an edit-like form, with the same validations:
  - if the value is empty, it fails `required` (`{ touched: true, dirty: false }`)
  - if the value is filled and it's not unique, it fails `unique` (`{ touched: true, dirty: true }`)
  - if the value is changed, but it's equal to the original value, do not fail `unique` (`{ touched: true, dirty: false }`) (does not work)
  - otherwise, valid (`{ touched: true, dirty: true }`)

It's kind of a poor example, but my point is: The validations API needs more context, having only the value + a set of values is not enough context for more specific validations, or to decouple validations, fields, components, and forms.

erikras:
Published fix in [`4.10.0`](https://github.com/final-form/final-form/releases/tag/v4.10.0).

PR #125, opened 2018-05-02
feature: Include field metadata as argument to field validators

gustavovnicius:
Fix for #124 

*Important*: Work in progress, opened the PR after spiking through `final-form`, so we can discuss and agree on a final and more refined solution.

### Overview
Pass the field metadata as a third argument to any validator function, enabling validators to also consider the field state (active, dirty, touched, etc...) while validating.

This is specially useful in cases where there is a validator composition since some validations require more context than the value.

Example: imagine an `edit` form sharing the same validations as the `create` form, in this form a `unique` validation for a previously filled field should only check for uniqueness if the value has changed. Validating if it hasn't changed will always fail the validation, and there's no way to tell today since there's no way to access a field's metadata (in this case, `dirty`). 

This will enable more generic fields and validators definition.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -270,8 +270,11 @@ const createForm = (config: Config): FormApi => {
         const errorOrPromise = validator(
           getIn(state.formState.values, field.name),
           state.formState.values,
-          publishFieldState(state.formState, state.fields[field.name])
+          validator.length === 3
+            ? publishFieldState(state.formState, state.fields[field.name])
+            : undefined
         )
+
         if (errorOrPromise && isPromise(errorOrPromise)) {
           const asyncValidationPromiseKey = nextAsyncValidationKey++
           const promise = errorOrPromise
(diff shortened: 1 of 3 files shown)


erikras:
Seems okay. I wonder if it could check `validator.length` first, to see if the function was expecting a third argument, so that it wouldn't have to go through `publishFieldState` every time for use cases not needing this functionality?

gustavovnicius:
@erikras Added the arity check, tried to look at adding a few tests before, not 100% sure though were to add them, can you give me some help?

erikras:
The coverage is all `100`s when you run `npm test`? If so, you're good.

gustavovnicius:
Yes it is, so I think we're good to go then :)

erikras:
You forgot the docs, but I'll get it. 👍

erikras:
Published in [`4.10.0`](https://github.com/final-form/final-form/releases/tag/v4.10.0).

Issue #102, opened 2018-04-03
Register fields without having to use the <Field /> component


dunika:
### Are you submitting a **bug report** or a **feature request**?

**feature request**

### What is the current behavior?

Currently it would seem that the only way to register a field is to define a named `<Field />` component inside a `<Form />` component. 

I have fields which I wish to update pragmatically. But it seems that I cannot change their values if they are not registered. For example I have something similar to what is below. I am passing `<LocationSearch/>` to `<FormSpy />` so I can access its the `change` prop. When I try to make the changes inside `componentDidUpfate` the fields are not updated. If I am to change the 'locationSearch' field though it will work which leads me to think that this is because it is a registered field.

```
class LocationSearch {
  async componentDidUpdate(prevProps) {
    const { change, input: { value }, values, mutators } = this.props;
    if (value && !shallowEqual(value, prevProps.input.value)) {
      const { address, location } = await getGooglePlaces().getPlaceDetails(value.id);

      // These changes won't work 
      change('location', location);      
      change('address', address);

      // But I if I am to do the following locationSearch will be updated 
      change('locationSearch', address);
      
    }
  }
}

const LocationSearchFormSpy = (props) => (
  <FormSpy
      component={LocationSearch}
  />
)

<Form 
  ...
  initialValues= {{
    address: {},
    location: null,
  }}
  render={({...}) => (
    <form>
      <Field
        name="locationSearch"
        component={LocationSearchFormSpy}
      />
      ...
      </form>
  )}
/>

```

I have tried using `mutators` also but have had no luck with them either. Maybe there is a better way of doing this that I am not aware of.

Anyway if you can help me out I'd greatly appreciate it. 

Great library btw!

erikras:
So you want all of [`Api`](https://github.com/final-form/final-form#api) available inside `<Form>`, not just the crippled version that is currently in [`FormRenderProps`](https://github.com/final-form/react-final-form#formrenderprops)?

dunika:
I'm able to get around this with the following for now 

`<Field name="location" component={() => null} />`

erikras:
Haha!! Long live the [`noop`](https://api.jquery.com/jQuery.noop/)!

erikras:
I'm going to close this because it can be achieved with a simple component.

```jsx
const NoopFields = ({ names }) => (
  <React.Fragment>
    {names.map(name => <Field name={name} render={() => {}} />)}
  </React.Fragment>
)

[...]

PR #134, opened 2018-06-08
Allow form.change() to set values for non-registered fields

erikras:
Fixes https://github.com/final-form/final-form-calculate/issues/15.
Fixes https://github.com/final-form/final-form/issues/102.

I have been thinking about why I originally chose to not allow changes to non-registered fields, and I cannot defend the decision. It makes the library less flexible. This PR lets you now modify field values for fields that are not currently registered.


erikras:
Published in [`v4.8.1`](https://github.com/final-form/final-form/releases/tag/v4.8.1).

PR #135, opened 2018-06-08
Use Object.keys instead Object.entries

Andarist:


erikras:
I always try to avoid that. Sadly, I wish I didn't have to. Looks like it was introduced in #123.

Thanks for keeping this lib in line for ye olde browsers. 👍

Andarist:
😉 happens to the best. I totally understand if u chose to drop support for older browsers in general (although I prefer u dont) but as long as u transpile syntax down to es5 I think the same rule should apply to using other javascript features.

Would u be able to do a release with it today?

erikras:
I was in the middle of deploying #134, so you _just_ made it.

Indiana

Andarist:
Super cool, thanks! ❤️

erikras:
Published in [`v4.8.1`](https://github.com/final-form/final-form/releases/tag/v4.8.1). Good save, Indy!

PR #136, opened 2018-06-12
Fix/improve TS declaration

milesj:
I noticed a handful of fields missing from the TypeScript declaration, so I updated it to match the Flow one.

In regards to TS, the type `object` isn't ideal, as it doesn't support nested objects correctly. It really needs to be something like `AnyObject`. Furthermore, I would highly suggest typing data/errors/other objects, for example:

```
export interface Data {
  [key: string]: any;
}

export interface Errors {
  [key: string]: string;
}
```

Seems like a bigger task, as Flow would need it to, so I opted not to do it.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -20,9 +20,11 @@ export interface FormSubscription {
   submitFailed?: boolean
   submitting?: boolean
   submitSucceeded?: boolean
+  touched?: boolean
   valid?: boolean
   validating?: boolean
   values?: boolean
+  visited?: boolean
 }
 
 export interface FormState {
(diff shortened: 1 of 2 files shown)


src/index.d.ts
```diff
@@ -53,7 +57,7 @@ export interface FieldState {
   active?: boolean
   blur: () => void
   change: (value: any) => void
-  data?: object
+  data?: AnyObject
```
milesj: I use nested objects here, which currently doesn't type correctly.

erikras:
Seems like an improvement. Thanks!

erikras:
Published in [`v4.8.2`](https://github.com/final-form/final-form/releases/tag/v4.8.2).

milesj:
Thank you!

PR #137, opened 2018-06-15
Fixed validation bug when changing unregistered field

erikras:
Added a check to make sure a field is registered when inspecting its `validationFields` property.

Fixes https://github.com/final-form/react-final-form/issues/278.

### DIFF
--- a/src/FinalForm.creation.test.js
+++ b/src/FinalForm.creation.test.js
@@ -34,4 +34,10 @@ describe('FinalForm.creation', () => {
     expect(form.getState().pristine).toBe(true)
     expect(form.getState().dirty).toBe(false)
   })
+
+  it('should allow a change to an not-yet-registered field when validation is present', () => {
+    const form = createForm({ onSubmit: onSubmitMock, validate: () => {} })
+    form.registerField('whatever', () => {}, { value: true })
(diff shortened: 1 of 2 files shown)


alexjsdev:
hey, I was refactoring some code based on what you said [here](https://github.com/final-form/final-form-calculate/issues/15)  and just ran into this bug, the strange thing is that: 
```
cat node_modules/final-form/package.json | grep version
 "version": "4.8.2",
```
but
 
```
cat node_modules/final-form/dist/final-form.cjs.js | grep version
var version = '4.8.1';
exports.version = version;
```

... seems like the CJS version didn't get built.... ¯\_(ツ)_/¯

alexjsdev:
aaaand after writing that just realized the version mismatch doesn't matter this is just not merged yet. I already made a PR to do (almost) exactly what you did, and then saw this, and thought it was merged since it was 4 days ago, and would be in 4.8.2... my bad on assuming that.

Anyways, can you please merge and release this? it helps me solve an autofill bug AND get rid of a bunch of hacky code i wrote. Appreciate it! Thanks for final form!

erikras:
Published in [`v4.8.3`](https://github.com/final-form/final-form/releases/tag/v4.8.3).

PR #150, opened 2018-07-31
Ensured calculated form state returned from getState()

erikras:
Addresses issue raised here: https://github.com/final-form/react-final-form/issues/310#issuecomment-409239501


erikras:
Published in [`v4.9.1`](https://github.com/final-form/final-form/releases/tag/v4.9.1).

PR #153, opened 2018-08-23
New mutator: renameField

erikras:
Handles moving over subscribers and values.


erikras:
Published in [`4.10.0`](https://github.com/final-form/final-form/releases/tag/v4.10.0).

PR #156, opened 2018-09-05
do not silently catch errors in promise in onSubmit

morloy:
Every error in an asynchronous `onSubmit` gets silently ignored at the moment.
Example:
`onSubmit={() => { throw new Error('test') }}` => Throws an error.
`onSubmit={async () => { throw new Error('test') }}` => Nothing happens.

From my understanding, this shouldn’t happen. If there’s a controlled submission error, one should call  something like `return { error: 'test'}` and not `throw`.
We‘re using Sentry to detect any uncaught exception in our production app, thus throwing errors is a critical feature for us.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -917,7 +917,7 @@ const createForm = (config: Config): FormApi => {
         if (result && isPromise(result)) {
           // onSubmit is async with a Promise
           notifyFormListeners() // let everyone know we are submitting
-          return result.then(complete, complete)
+          return result.then(complete)
         } else {
           // onSubmit is sync
           complete(result)


erikras:
This line was changed in #148 to allow for `submitting` and `validating` flags to flip to false on `Promise` rejection.

What if it _still_ ran `complete()`, but then also threw the error again, I wonder... 🤔

morloy:
Personally, I assume an uncaught exception is a critical error. Thus, I wouldn’t care about `submitting` or `validating` flags afterwards, the user has to reload the page anyways. Nonetheless, if we could have both, that’s also fine.

erikras:
Published fix in [`v4.11.0`](https://github.com/final-form/final-form/releases/tag/v4.11.0).

PR #165, opened 2018-10-03
Rely less on onSubmit.length to determine callback style

jorrit:
This PR fixes a potential time waste for future programmers that accidentally have the following `onSubmit` function:
```js
handleSubmit = (values: object, form: FormApi, callback?: (errors?: object) => void) => {
  return new Promise(....);
});
```

The current code first checks if the number of parameters is 3 and then assumes that the onSubmit function uses the callback style. My suggestion is to first check if the function returns a promise and assume Promise style if this is the case. Another solution would be to enhance the documentation and add a warning that Promise-based submit handlers should never have three parameters.


erikras:
Good thinking. 👍

jorrit:
You're the best, thanks!

PR #167, opened 2018-10-10
Use @babel/runtime helpers instead of bundled ones

Andarist:
Same as https://github.com/final-form/react-final-form/pull/351


erikras:

PR #170, opened 2018-10-18
Rethrow submit errors so promise rejects

erikras:
Closes #156.

Also updated deps.


erikras:

PR #171, opened 2018-10-26
Add TypeScript typings to form data values

bwlt:
`createForm` now have a type parameter.

Sorry for the reformatting on `src/index.d.test.ts`, but it was caused by the precommit hook


erikras:

PR #173, opened 2018-10-30
Add `submitting` to FieldState

### LINKED ISSUE #172: Add `submitting` to `FieldState`
That way it would be easier to disable a field while the form is being submitted 

I currently have a custom component that takes care of nicely displaying errors etc, so my code looks like this:

```jsx
<Field name="foo" component={MyFancyFormField} as={TextArea}>
```

`MyFancyFormField` is basically this:

```jsx
export function MyFancyFormField(
{
    input, label, placeholder, required, children, disabled, componentLabel, defaultValue,
    meta: {touched, error, submitError, submitting, dirty, dirtySinceLastSubmit},
    as: Component,
    ...props
}
) {
    // we show errors if:
    // - the field was touched (focused+unfocused)
    //   ...and failed local validation
    //   ...and does not have the initial value
    // - there was an error during submission
    //   ...and the field has not been modified since the failed submission
    let errorLabel = null;
    if (touched && error && (dirty || required)) {
        errorLabel = <Label basic color="red" pointing="above" content={error} />;
    } else if (submitError && !dirtySinceLastSubmit) {
        errorLabel = <Label basic color="red" pointing="above" content={submitError} />;
    }

    return (
        <SUIForm.Field required={required} error={!!errorLabel} defaultValue={defaultValue}>
            {label && <label>{label}</label>}
            <Component {...input}
                       {...props}
                       label={componentLabel}
                       placeholder={placeholder}
                       disabled={disabled || submitting} />
            {errorLabel}
            {children}
        </SUIForm.Field>
    );
}
```

Currently the `disabled={disabled || submitting}` logic doesn't work, because `meta.submitting` doesn't exist (iirc it's a leftover from when we used redux-form which [did include it there](https://redux-form.com/7.4.2/docs/api/field.md/#-code-meta-submitting-boolean-code-))

ThiefMaster:
This was present in redux-form but not in final-form.

closes #172


erikras:

PR #176, opened 2018-11-05
Remove undefined values rather than setting them to undefined

Andarist:
Seems to me like a better logic for this use case


erikras:
I like it.

erikras:

PR #181, opened 2018-11-15
export runValidation api

493636333:
hello,i have some requirement, sometimes i have to runValidation do myself, but now the api is not export.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "4.5kB"
+      "maxSize": "4.6kB"
     },
     {
       "path": "dist/final-form.es.js",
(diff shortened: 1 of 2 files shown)


erikras:
You should never need to do this. Validation is run on every change. Why would you need this?

493636333:
I know validation is run every chang,but if validator is changed,validation will not run

493636333:
if I can do runValidation do myself? because I want run validation when other field changed or validator changed, or one of props changed, not only current field value changed

erikras:
Can you describe your use case a little better? Which validation function is changing? Why is it changing?

493636333:
i have a requirement, for example, there have two fields, one is a select field(a), one is another custom component(b), there will fetch some data(sourceData) from api when a is changed, then validation of b field depend on sourceData, but now i can't call validate when sourceData is back。

erikras:
Could you not use async validation (i.e. return a `Promise` from your validation function)?

493636333:
no，async data not only use in validation，custom Field need this async data，if i can run validation do myself，i think everything will be resolved。

PR #190, opened 2018-12-13
changes to available version of event-stream in package-lock

petegleeson:
Hi 👋

I was running into a 404 problem while doing an `npm install`. A bit of googling turned up [this result](https://stackoverflow.com/questions/53578201/npm-err-404-not-found-event-stream3-3-6). It seems like there was a malicious version of `event-stream` published and that version is not available anymore. I manually modified the package-lock to specify the patch before the bad version.


erikras:
Oops. Did this in #197 before seeing this. Thanks!

Issue #195, opened 2019-01-31
Canges on `formState.submitting` not being published timely


nicooga:
### Are you submitting a **bug report** or a **feature request**?

This is a bug report (?)

### What is the current behavior?

Changes on `fieldState.submitting` are not being published when I'd expect them to.
Changes on `formState.submitting` **are being published**, but using form subscriptions instead of field subscriptions in order to get notified when form is submitted is sub-optimal.

### What is the expected behavior?

I expect for subscriptors `fieldState.submitting` to be updated inmediately when form begins to submit asynchronously.

### Sandbox Link

https://codesandbox.io/s/n0oq6x95jm

### What's your environment?

Not so relevant I think, as I was able to reproduce the error in a plain sandbox.
This should also not be relevant but I first found this problem while working with `react-final-form`.
I've tried to reproduce the error in plain Vanilla JS to discard the possibility that the wrapper package is guilty.

final-form: 4.11.0
browser/OS: Chromium version 71.0.3578.98 (Official Build) Built on Ubuntu , running on Ubuntu 18.04 (64-bit)

gbiryukov:
Facing the same issue. PR #196 is really appreciated

erikras:
Published fix in [`v4.11.1`](https://github.com/final-form/final-form/releases/tag/v4.11.1).

PR #196, opened 2019-01-31
Fix fields not being notified about form being submitted

nicooga:
Fixes https://github.com/final-form/final-form/issues/195.


erikras:
Published in [`v4.11.1`](https://github.com/final-form/final-form/releases/tag/v4.11.1).

PR #197, opened 2019-02-06
Allow array items to optionally be destroyed in setIn()

erikras:
Fixes https://github.com/final-form/react-final-form-arrays/issues/35.


erikras:
Published in [`v4.11.1`](https://github.com/final-form/final-form/releases/tag/v4.11.1).

PR #199, opened 2019-02-07
accept function as argument in initialize()

alirezavalizade:
Usage:
```
initialize((values) => ({ ...values, any: null }))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ form.submit() // only submits if all validation passes
     - [`getFieldState: (field: string) => ?FieldState`](#getfieldstate-field-string--fieldstate)
     - [`getRegisteredFields: () => string[]`](#getregisteredfields---string)
     - [`getState: () => FormState`](#getstate---formstate)
-    - [`initialize: (values: Object) => void`](#initialize-values-object--void)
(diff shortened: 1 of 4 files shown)


alirezavalizade:
If you merge this, I'll send a pull request for `react-final-form` docs.
@erikras

erikras:
Can you describe the use case for this?

alirezavalizade:
Sure, Sometimes i need to change value of a field but also need to keep the current values of form,
for example: 

```
initialize({
  posts: 'a value'
})
```
If i use this other fields will lose their values. But i need to keep my current values and just change one field. and also i don't want to set values in state on changing the form. I know it's possible using `change` method in `finalForm` state but sometimes need to this. It's nice to have it.
[Take a look at this sandbox please](https://codesandbox.io/s/qm8y71wzw)

erikras:
Well, this could be accomplished like this:

[Edit 🏁 React Final Form - Loading and Initializing](https://codesandbox.io/s/03kv4mpwml)

You have all the information you need to initialize with all the values.

alirezavalizade:
@erikras Thanks for your reply, Yes i agree it, but really sometimes need to access current values for initialize the form. I can create another Sandbox,

PR #204, opened 2019-03-13
adding formName to formApi

vospascal:
I would like to add support for name property/attribute to final-form since form name is a default property/attribute, its handy to get some context when dynamically and re-using custom field components. 
This is part one of a two part pull request the other one will be for react-final-form.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     }
   ],
   "dependencies": {
-    "@babel/runtime": "^7.3.1",
-    "settings": "^0.1.1"
+    "@babel/runtime": "^7.3.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "4.8kB"
+      "maxSize": "4.9kB"
     },
     {
       "path": "dist/final-form.es.js",
@@ -99,6 +99,7 @@
     }
   ],
   "dependencies": {
-    "@babel/runtime": "^7.3.1"
+    "@babel/runtime": "^7.3.1",
+    "settings": "^0.1.1"
   }
 }
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -28,7 +28,7 @@ export interface FormSubscription {
   validating?: boolean
   values?: boolean
   visited?: boolean
-  formName?: string
+  formName?: boolean
 }
 
 export interface FormState {
@@ -105,7 +105,7 @@ export interface FieldSubscription {
   valid?: boolean
   value?: boolean
   visited?: boolean
-  formName?: string
+  formName?: boolean
 }
 
 export type FieldSubscriber = Subscriber<FieldState>
@@ -265,7 +265,6 @@ export interface Config<FormData = object> {
   validate?: (values: object) => object | Promise<object>
   validateOnBlur?: boolean
   name?: string
-  formName?: string
 }
 
 export type Decorator = (form: FormApi) => Unsubscribe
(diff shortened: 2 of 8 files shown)


vospascal:
i dont know wy travis is failing on some node versions could use some help there

package.json
```diff
@@ -99,6 +99,7 @@
     }
   ],
   "dependencies": {
-    "@babel/runtime": "^7.3.1"
+    "@babel/runtime": "^7.3.1",
+    "settings": "^0.1.1"
```
Sharakai: What's this `settings` dependency for? I can't see it being used anywhere.

src/index.d.ts
```diff
@@ -102,6 +105,7 @@ export interface FieldSubscription {
   valid?: boolean
   value?: boolean
   visited?: boolean
+  formName?: string
```
Sharakai: This should be a `boolean`

src/index.d.ts
```diff
@@ -256,6 +264,8 @@ export interface Config<FormData = object> {
   ) => object | Promise<object | undefined> | undefined | void
   validate?: (values: object) => object | Promise<object>
   validateOnBlur?: boolean
+  name?: string
+  formName?: string
```
Sharakai: `formName` doesn't seem to be a valid config key.
You're using `name` in the form config, and aliasing it to `formName`.
Only `name` should exist here.

src/types.js.flow
```diff
@@ -251,14 +256,17 @@ export type Config = {
...
   mutators?: { [string]: Mutator },
   onSubmit: (
     values: Object,
     form: FormApi,
     callback: ?(errors: ?Object) => ?Object
   ) => ?Object | Promise<?Object> | void,
   validate?: (values: Object) => Object | Promise<Object>,
-  validateOnBlur?: boolean
+  validateOnBlur?: boolean,
+  name?: string,
```
Sharakai: Same here regarding the `name` and `formName` mix-up. Only `name` is a real config property.

vospascal:
@Sharakai thnx will fix

vospascal:
@Sharakai you got time to check out https://github.com/final-form/react-final-form/pull/443 this implements the formName

vospascal:
@erikras do you have anytime soon to merge the requests or have a discussion about adding it?
If you need some help with the awesome library i'm willing to where possible.

erikras:
I'm not sure I understand why Final Form needs to maintain an additional superfluous string that doesn't change. Can you explain it further?

vospascal:
@erikras @Sharakai Thats easy this is the way to add context and pass it down while maintaining re-uasbility for instance with validation form's are the containers that hold the fields the fields and for example validation can be re-used over multiple forms. In for any component based architecture react/angular/vue this will be very helpfull. This way you dont have to resort to property drilling. Thats wy i added also an pull request to the react-final-form repo.

erikras:
Can you give me a concrete example? I'm still not seeing the utility.

vospascal:
The answer is context without property drilling.
Having a static named property in any case makes configuration of reusable components doable.
It could also be something else then the form name aslong as it provides context to the fields inside the form. 
I know this is wat makes it look like its superfluous, redundant, unneeded but its the only property that provides basic context. 
Any other suggestions how to achieve this are welcome.

erikras:
_Who_ needs to know the form name? The validation function? The custom input component?

vospascal:
@erikras that should also work indeed good idea! That i didnt think of that before.... 
Dont know if that also goes for the non react usages of final-form but this will work fine for my usecase i will implement this first thing tomorrow.

vospascal:
Got it implemented looks like it works prefectly thnx ill close this issue.
I also got an implementation using a decorator for meta data that might be interessting. ill see if i got the time this week to wip up an example implementing an typeahead.
I hope you can take a look.

Issue #193, opened 2019-01-10
setInRecursor extremely slow


bertho-zero:
### **Bug report**

I have a big form (8000 checkboxes) with final-form-calculate and when I check a box it can touch up to 600 other checkboxes, everything works perfectly, but it's very slow.
I did everything to optimize the thing, but a big slowness persists, and it comes from `setInRecursor`.

I have neither validation nor anything special, and when I uncheck a box that uncheck 600 others, it takes **3.92s**!

### What's your environment?

Latest 🏁 Final Form and Node versions, Chrome on Linux.

### Other information

![final-form_issue](https://user-images.githubusercontent.com/8525267/50964188-e6d44400-14ce-11e9-97cd-f6eef81da4f0.png)

bertho-zero:
@erikras Here are the expensive lines, it represents 3.18 / 3.60s

![image](https://user-images.githubusercontent.com/8525267/55716704-f1cc4100-59f7-11e9-90c1-32628687a9a0.png)

erikras:
Published possible fix in [`v4.13.0`](https://github.com/final-form/final-form/releases/tag/v4.13.0).

bertho-zero:
It is particularly slow if you have a big form with a mutator that causes value changes.  The object is copied completely multiple times.

PR #211, opened 2019-04-08
improve performance of setIn

bertho-zero:
`Object.keys` is called only when necessary.

I go from 3.57s to 2.93s by making a change on a huge form. (related to https://github.com/final-form/final-form/issues/193)

The remaining slow part is:
https://github.com/final-form/final-form/blob/f02fd2699e4cb056b9d55882c1ac0d80e77ef9e7/src/structure/setIn.js#L66-L69

### DIFF
--- a/src/structure/setIn.js
+++ b/src/structure/setIn.js
@@ -44,8 +44,8 @@ const setInRecursor = (
       value,
       destroyArrays
     )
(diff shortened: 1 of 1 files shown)


erikras:
Published in [`v4.13.0`](https://github.com/final-form/final-form/releases/tag/v4.13.0).

PR #214, opened 2019-05-06
guard against deleted entry in notify

onlyann:
When final form notifies form listeners on `complete`, it can happen that notifying one subscriber will cause another subscriber to unsubscribe, effectively removing the entry from the iterated list.

The [following example](https://codesandbox.io/s/zwqxoz8nn3) shows how it can throw:
```
final-form.cjs.js:404 Uncaught (in promise) TypeError: Cannot read property 'subscription' of undefined
    at eval (VM20664 final-form.cjs.js:404)
    at Array.forEach (<anonymous>)
    at notify (VM20664 final-form.cjs.js:402)
    at notifyFormListeners (VM20664 final-form.cjs.js:824)
    at complete (VM20664 final-form.cjs.js:1179)
```

The PR adds a guard to prevent the error.

Fixes https://github.com/final-form/react-final-form/issues/159

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -133,8 +133,11 @@ function notify<T: Object>(
   filter: StateFilter<T>
 ): void {
   Object.keys(entries).forEach(key => {
-    const { subscription, subscriber } = entries[Number(key)]
-    notifySubscriber(subscriber, subscription, state, lastState, filter)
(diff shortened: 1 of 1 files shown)


erikras:
Published in [`v4.13.0`](https://github.com/final-form/final-form/releases/tag/v4.13.0).

PR #218, opened 2019-05-17
Improve mount/unmount performance

docentq2:
Hi!
We use final-form in our product to display    a large number of fields (over 500). When we switch between forms, we wait a very long time when the form is replaced.
I found out that performance is greatly increased if you use a more efficient regexp in the toPath method.


docentq2:
Unfortunately, the performance boost appeared only in the development environment. In production mode it is not :(

PR #224, opened 2019-06-11
Update types.js.flow

morloy:
It feels like `initialValues` and `onSubmit` should receive the type as well.


erikras:
Oversight on my part. Thanks!

erikras:
Published in [`v4.14.1`](https://github.com/final-form/final-form/releases/tag/v4.14.1).

PR #229, opened 2019-06-14
Throw error if attempt to reset() when submitting

### LINKED ISSUE #142: Calling `reset()` in onSubmit results in inconsistent state
https://codesandbox.io/s/mzm3m87vvx

When I `reset()` at the end of the (async) onSubmit function, the fields are reset, but the submitSucceeded flag is not reset. When calling reset() at some other point (e.g. the reset button, or even with in a zero-second setTimeout started in onSubmit), it works fine.

erikras:
Fixes #142.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ✅ Opt-in subscriptions - only update on the state you need!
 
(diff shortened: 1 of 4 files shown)


erikras:

PR #230, opened 2019-06-14
Lookup version at build time

erikras:
Like https://github.com/final-form/react-final-form/pull/532


erikras:

PR #232, opened 2019-06-16
Added a validating key  to fieldState, will be true when promise returned from a field validate function,useful for individual file uploads in form.

contactyash:
Sometimes when uploading files directly from input on change, instead of on submit, in that case that Field needs to provide a key like ```validating``` equals to true  till promise resolves, to show loader.

Every Field providing ```subscribeToEachFieldsPromise``` equals to true will receive notification every time validate function's promise is resolves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,15 +88,15 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "4.9kB"
+      "maxSize": "4.94kB"
     },
     {
       "path": "dist/final-form.es.js",
-      "maxSize": "8.5kB"
+      "maxSize": "8.52kB"
     },
     {
       "path": "dist/final-form.cjs.js",
-      "maxSize": "8.5kB"
+      "maxSize": "8.65kB"
     }
   ],
   "dependencies": {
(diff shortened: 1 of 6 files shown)


erikras:
Overall this looks good. People have been requesting a per-field `validating` flag for a while.

I disagree with the need for a `subscribeToEachFieldsPromise` config parameter. Seems like if you're using async field-level validation, you should get this functionality automatically. I think the damage that option causes to the API surface area is worse than the `shallowEqual` call it was avoiding.

This needs a little clean-up, typescript types, and docs, but I'll handle that in a subsequent PR.

Nice work! 👍 That bit you edited is the most complicated part of the library.

erikras:
Upon further thought, it doesn't make sense to keep a separate map for `fieldsWithPromisePending`. That `boolean` flag should just go right into the `InternalFieldState`. I've reimplemented it this way, keeping your same, excellent tests, and it works great, with a slightly smaller bundle size. See #235.

Thank you for causing me to think this through. 👍

erikras:
Published fix in [`v4.16.0`](https://github.com/final-form/final-form/releases/tag/v4.16.0).

contactyash:
Thank you for motivating words. This was my first pull request.
One question I have is now we are not calling runValidation function's callback on each promise resolved. Which I guess just means we have provided form's validating to each field also. It will stop all field's loaders at once when all promises resolved. And guess test are passing because we are subscribed to just one field, when subscribed to field resolving at different times will need more test cases.

PR #235, opened 2019-06-19
Field validating flag

erikras:
Reimplements #232 keeping `validating` state directly in `InternalFieldState`.


erikras:

PR #236, opened 2019-06-22
Clean submitError & submitErrors when submitting

### LINKED ISSUE #226: Stale submit errors might get shown for async submits

bug report~

### Description

If we fill in an invalid value, submit it and receive server-side error regarding the incorrect value, then correct ourselves and submit again we are going to see that previous/stale error being shown.

This happens because `dirtySinceLastSubmit` is being set synchronously to `false` for async submit (actually it's derived from `lastSubmittedValues`) AND `submitErrors` errors are being updated only AFTER submit resolves.

So when performing a second "corrected" submit we land briefly in a state when we still have stale submitError in a form state and when final form considers the form as "not dirty" (because current values match `lastSubmittedValues`).

I'd like to fix it in the core, but I'm wondering what's the exact expected behaviour, we can either:
- "commit" `lastSubmittedValues` when submit resolves, not when it gets called
- clear submitErrors & submitError before actual onSubmit gets called
- or maybe even both? 🤔 

Submit error being shown when submitting.

No stale error being shown when submitting

### Sandbox Link

https://codesandbox.io/s/weathered-hill-omtok

Andarist:
fixes #226

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -844,9 +844,7 @@ function createForm<FormValues: FormValuesShape>(
 
     reset: (initialValues = state.formState.initialValues) => {
       if (state.formState.submitting) {
-        throw Error(
-          'Cannot reset() in onSubmit(), use setTimeout(form.reset)'
-        )
+        throw Error('Cannot reset() in onSubmit(), use setTimeout(form.reset)')
       }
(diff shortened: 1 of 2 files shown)


erikras:

PR #237, opened 2019-06-24
Fixed TS GetFieldValidator

s97712:


erikras:
Can you explain why, if `getValidator` is [already optional](https://github.com/final-form/final-form/blob/master/src/index.d.ts#L133)?

erikras:
Doesn't hurt, I guess...

erikras:

Issue #240, opened 2019-07-09
Mutators not working as expected


kaidjohnson:
### Are you submitting a **bug report** or a **feature request**?

Bug Report

### What is the current behavior?

Triggering a mutator on a field before it is registered results in the mutation not being set. Triggering a mutator after the field has been registered results the first subscriber callback to not have the mutation, but subsequent callbacks to have it.

This behavior is expected given how mutators work, however the documentation is misleading.

Given the example from https://github.com/final-form/final-form-set-field-data

```
form.mutators.setFieldData('firstName', { awesome: true })

form.registerField(
  'firstName',
  fieldState => {
    const { awesome } = fieldState.data // true
  },
  {
    // ...other subscription items
    data: true
  }
)
```

We're expecting `awesome` to be `true`, however, it is `undefined`, as seen here:

https://codepen.io/ThunderheadONE/pen/NZEBGe?editors=1112

If we move the mutator to after the registerField, we get one churn with `undefined` (before the mutator), followed by a second update (triggered by the mutator itself) with the expected value of `true`.

This issue might be related to [#178]

### What is the expected behavior?

At the very least, the documentation probably needs to be updated with a working example and potentially get a few additional complex examples added. 

### Sandbox Link

https://codepen.io/ThunderheadONE/pen/NZEBGe?editors=1112

Syynth:
I think this is probably just a documentation issue around `final-form-set-field-data` as the entire implementation of that mutator is just
```javascript
const setFieldData: Mutator = (args: any[], state: MutableState) => {
  const [name, data] = args
  const field = state.fields[name]
  if (field) {
    field.data = { ...field.data, ...data }
  }
}
```
so it looks pretty intentional that it's dropping the data because the field has not yet been registered. I suppose you could create your own field information on the state, but the logic is quite complex (relatively speaking) in that it is over 100 lines and manages field subscribers and validation setup that the form needs.

A very simple change that could be made would be to extend the `formApi.registerField`'s 4th argument (`fieldConfig`) to also accept a `data` attribute. This would actually be a one line change at [L764](https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L764), changing `data: {}` to `data: fieldConfig && fieldConfig.data || {}`.

erikras:
Published @Syynth's fix in [`v4.16.2`](https://github.com/final-form/final-form/releases/tag/v4.16.2). You can now provide an initial `data` value when you register your field.

PR #241, opened 2019-07-10
Added data to FieldConfig

Syynth:
This change would allow `form.registerField` to initialize the `data` property for the field being registered, like `initialValue`, `defaultValue`, etc. I implemented this in response to seeing #240. Hopefully I've covered all my bases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,6 +530,10 @@ A callback to notify fields after submission has completed successfully.
 
 A function to call just before calling `onSubmit`. If `beforeSubmit` returns `false`, the submission will be aborted. If one of your fields returns `false` on `beforeSubmit`, other fields may not have their `beforeSubmit` called, as the submission is aborted on the first one that returns `false`.
 
+#### `data?: any`
+
+Initial state for arbitrary values to be placed by mutators.
+
 #### `defaultValue?: any`
 
 ⚠️ You probably want `initialValue`! ⚠️
--- a/src/FinalForm.creation.test.js
+++ b/src/FinalForm.creation.test.js
@@ -114,4 +114,35 @@ describe('FinalForm.creation', () => {
       pristine: false
     })
   })
+
+  it('should allow data to come from field when registered', () => {
+    const form = createForm({ onSubmit: onSubmitMock })
+    const foo = jest.fn()
+    const cat = jest.fn()
+    form.registerField(
+      'foo',
+      foo,
+      { pristine: true, data: true },
+      { data: { foo: 'bar' } }
+    )
+    expect(form.getFieldState('foo').data).toEqual({ foo: 'bar' })
+    form.registerField(
+      'cat',
+      cat,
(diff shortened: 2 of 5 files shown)


erikras:
This looks great. 👍

Out of curiosity, what sort of "arbitrary data" are you storing with your fields?

erikras:

bertho-zero:
@erikras An `intermediate` state for the checkboxes, for example.

Syynth:
@erikras I am using it for a few things. One is storing extra validation metadata, so that I can generate rich "checklists" for filling out the form that can gives users additional context w.r.t. field errors and so on.

One additional thing I'm doing that I would actually be willing to try another PR for is having a matching `data` field on the top level `formState` so I can store metadata related to the entire form record.

My use case for this is related to persisting form state, I'm currently using a decorator to write formState.values to sessionStorage/localStorage on any change, and then reloading from sessionStorage if the user navigates away and returns, but when exiting the page and re-entering, the data from localStorage is stored in the `data` portion of a special "__meta__" field, and when that is present a prompt is displayed at the top saying "there is old data, would you like to continue from where you left off?".

This is also used while editing CRUD forms with GraphQL subscriptions. If another user edits the same record, the new copy is stored in __meta__ and a separate prompt is displayed saying "this record has been updated since you began editing, would you like to reload the data?" or something similar. I store this data in the fieldState.data of __meta__ so it's not written to the localStorage cache as well, since that would create nested copies of cached data on each page refresh, for instance.

It works fine to use __meta__ as a dummy field for this but it would be nice to just have a top level place to use for this. I know I'm definitely stress-testing this library though (it holds up!) :)

erikras:
Thanks for that, guys!

Re: form-level `data`... My feeling was always that, at the top level, one could just use component state. The `data` thing was always just a little hack to use the existing structure of fields. [Shout out to jQuery!](https://api.jquery.com/jQuery.data/) At the "form level", you're not taking advantage of the tree of fields anymore. Before you submit a PR, try just using React to hold that state.

Syynth:
That's a totally valid point, I think, but my issue comes from wanting to work with that data in mutators. top level data isn't the only solution of course, but if there was a way to allow extra data or props or something can be passed along to mutators I think that would be very useful. One question that might mitigate this: does formApi.setConfig('mutators', ...) cause any kind of subscription notifications? If not I guess I could just use curried functions for mutators and re initialize the mutators when the component state changes, but that still requires my top level form component to re-render (which _hopefully_ doesn't cause the fields to re-render and come with a perf cost, but I don't know for sure some of our forms won't be implemented in a fashion that invalidates this concern.)

But this is going a bit long, should I open a new issue to discuss this? Or will my re-setting mutators solution be fairly performant? In any case, having a special `__meta__` field is totally sufficient for my use case currently. I really love this library and the philosophy of extensibility over lots of pre-included features so I do my best to keep that in mind before suggesting things

PR #243, opened 2019-07-11
Fixed bug in hasAnyError

### LINKED ISSUE #179: submitErrors object containing only Error instance is not considered an error

Bug; in my opinion.

 If returning `Error` or object having `Error` as value from `onSubmit` it is detected as not having error.

### Sandbox Link
https://codesandbox.io/s/93m26zqjrr

### What's your environment?
```
 "react-final-form": "^3.6.5",
  "final-form": "^4.10.0",
```

### Other information

This worked in
```
"react-final-form": "2.2.0",
"final-form": "3.0.1",
```

The culprit seems to be:
https://github.com/final-form/final-form/blob/42516e0f15430d0c894d775bc66b731cecd423aa/src/FinalForm.js#L63

A fix along the lines `if (value instance Error) return true` would be sufficient.

erikras:
Fixes #179.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "4.90kB"
+      "maxSize": "5.0kB"
     },
     {
       "path": "dist/final-form.es.js",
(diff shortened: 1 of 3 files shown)


erikras:

PR #244, opened 2019-07-13
Added internal forceUpdate flag for fields for final-form-arrays' mutators

erikras:
Mutators can now mark a field as `forceUpdate` to ensure that its subscribers will be notified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,8 +100,6 @@
     }
   ],
   "dependencies": {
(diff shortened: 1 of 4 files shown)


erikras:
Reverted in [`v4.18.0`](https://github.com/final-form/final-form/releases/tag/v4.18.0).

PR #246, opened 2019-07-15
Fix bug when mutator removes field state

erikras:


erikras:
Published in [`v4.18.1`](https://github.com/final-form/final-form/releases/tag/v4.18.1).

mohsinulhaq:
I was about to log an issue regarding this and I came across this commit at the exact line where I was getting the error at :)

Sharakai:
FWIW: this can also occur on the line below, with `delete state.fieldSubscribers[name].entries[index]` - Would probably be nice to extend this check to include `if (state.fieldSubscribers[name])` too.

PR #254, opened 2019-07-30
Improve ARRAY_ERROR & FORM_ERROR TypeScript types

danielnixon:


erikras:
This is undoubtedly _more true_, but I'm curious if it was causing a problem the old way?

danielnixon:
> This is undoubtedly _more true_, but I'm curious if it was causing a problem the old way?

Not really a problem _per se_, but with the more strict typing here it becomes easier to enforce this in types:

> Validation errors must be in the same shape as the values of the form. You may return a generic error for the whole form using the special FORM_ERROR string key.

PR #259, opened 2019-08-06
Improve `setConfig` TypeScript typing

danielnixon:


erikras:
Whoa! That's awesome. 👍

PR #260, opened 2019-08-06
Better compatibility with IE11

smashercosmo:
Array.prototype.find is not supported by IE11 and also seems like Array.prototype.some is more suitable for this case.


erikras:
I always forget about that. Thanks!

smashercosmo:
Thx for the lib)

Issue #261, opened 2019-08-07
[Feature Request] specified values subscription


kimochg:
### Are you submitting a **bug report** or a **feature request**?

feature request

### What is the current behavior?

```
subscription={ values: true }
```

### What is the expected behavior?

```
subscription={ values: { fieldA: true, fieldB: true } }
```

leads to notify in a more specified condition when fieldA and fieldB changed. and, not break current behaviour

### Other information

After checking the code, I think we can pass a `{ [fieldName: string]: boolean }` object as `subscription[values`

https://github.com/final-form/final-form/blob/59c23c82bff3c1d7f4fc5a47a1732d80442eb683/src/subscriptionFilter.js#L19

related issues: https://github.com/final-form/final-form/issues/105, https://github.com/final-form/final-form/issues/194, https://github.com/final-form/final-form/issues/177

valerybugakov:
Would be handy to have this feature. We're struggling with performance optimizations. It would make them more straightforward.

nvaken:
Same here, we have a huge form that has a sidebar with a table of contents, which needs to change per item when its corresponding field has changed. Subscribing to just "values" is a pretty big performance hog, rerendering the entire table of contents when just one field is changed.

PR #264, opened 2019-08-08
:sparkles: subscription shape

kimochg:
This PR supports object subscription for `values` and `value`.

For this issue https://github.com/final-form/final-form/issues/261


erikras:
Add a test showing that this works?

kimochg:
hi @erikras , codecov backs to 100% after adding some tests, plz take another look, thx.

chengcyber:
ping @erikras

Dragomir-Ivanov:
@erikras What can this be merged? I think that its valuable feature, and I could make use of it.

erikras:
If I'm understanding the desire correctly, it's that you want to pass something like:

```js
{
  values: {
    firstName: true,
    lastName: true
  }
}
```

...and _only_ be notified when the values of those two fields change? Sort of like a [MongoDB projection](https://docs.mongodb.com/manual/reference/operator/aggregation/project/index.html).

I've written a test that actually goes all the way through `createForm()` and it doesn't seem to work. Can you investigate further into why?

chengcyber:
Hi, @erikras.
After check the projection doc and the test case, I think there might be some concept issue here. The core part of my proposal is the projection of subscription to reduce the number of subscriber be called. I would like use the failing test to express my thoughts.
```
    form.subscribe(spy, {
      values: {
        name: true,
        email: true
      }
    })
```

means "I" only care about `name` and `email`. No matter what other values change won't notify this subscriber. Meanwhile the values of form data is not projected right now. if the form data contains `name`, `email`, `phone`, they are all passed to the component.

So, the failing test should be

```
    expect(spy.mock.calls[2][0]).toEqual({
      values: { name: 'erikras', email: 'erikras@final-form.org', phone: '867-5309' }
    })
```

which should be passed

the working part is 

```
    form.change('phone', '867-5309')
    expect(spy).toHaveBeenCalledTimes(2)
```

even the phone field changes, this subscriber not care about it, so no notification comes!

____

I'm not sure whether we should do form values projection. because it is a performance boost in my opinion. Any suggestions?

Dragomir-Ivanov:
The way to do it with recent React, is to create a component and repeat `useField("field-name"), with the fields you want to react to.

PR #265, opened 2019-08-09
feat: ignore old async-validation results

### LINKED ISSUE #147: Old/slow validate promises override errors from new/fast promises

An older, longer running validation promise will override the error state from a newer, shorter running validation promise.

Older validation promises are ignored and only the latest validation result is used.

### Sandbox Link

https://codesandbox.io/s/wy7z7q5zx5

In the username field, quickly enter any character and then delete it before the async validation can finish running. Notice the `Required` error briefly appears before being overridden by the async validation.

shrugs:
This PR introduces `Rx#switchMap`-like behavior into final-form async record- and field-level validations using the existing `nextAsyncValidationKey` logic, and applying it to the set of validations rather than individual validations. Then, when an async validation is resolved, we determine whether or not it is the most recent validation, ignoring it otherwise.

see https://github.com/final-form/final-form/issues/147#issuecomment-519881470 for discussion, but the gist is that when validation takes a conditionally long time to return, old values can take longer than new values (especially if your validation hits a server only _after_ sync validation passes) and we can enter invalid states when the old validation results overwrite the state after the latest validation result has already finished resolving.

I've included a test (and naturally the rest of the suite passes), but please check my work, since I'm less familiar with the library internals as a whole.

~One concern: the same test, applied to final-form#master, throws at the `toHaveBeenCalledTimes` because the `error` subscription spy has actually been called 3 times, despite there being only 2 error-triggering validations. The rest of the suite passes as expected. This may point to another, related bug that I've perhaps accidentally fixed within this PR, but I can't really reason well about it without spending too much time internalizing the library.~ nvm this is definitely because we're ignoring the old validation errors now, so one less render is triggered 😅 

Fixes #147

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -287,12 +287,7 @@ function createForm<FormValues: FormValuesShape>(
     if (validate) {
       const errorsOrPromise = validate({ ...state.formState.values }) // clone to avoid writing
       if (isPromise(errorsOrPromise)) {
-        const asyncValidationPromiseKey = nextAsyncValidationKey++
-        const promise = errorsOrPromise
-          .then(setErrors)
-          .then(clearAsyncValidationPromise(asyncValidationPromiseKey))
-        promises.push(promise)
-        asyncValidationPromises[asyncValidationPromiseKey] = promise
+        promises.push(errorsOrPromise.then(setErrors))
       } else {
         setErrors(errorsOrPromise)
       }
@@ -327,16 +322,12 @@ function createForm<FormValues: FormValuesShape>(
         )
 
         if (errorOrPromise && isPromise(errorOrPromise)) {
-          const asyncValidationPromiseKey = nextAsyncValidationKey++
           field.validating = true
-          const promise = errorOrPromise
-            .then(error => {
-              field.validating = false
-              setError(error)
-            }) // errors must be resolved, not rejected
-            .then(clearAsyncValidationPromise(asyncValidationPromiseKey))
+          const promise = errorOrPromise.then(error => {
+            field.validating = false
+            setError(error)
+          }) // errors must be resolved, not rejected
           promises.push(promise)
-          asyncValidationPromises[asyncValidationPromiseKey] = promise
         } else if (!error) {
           // first registered validator wins
           error = errorOrPromise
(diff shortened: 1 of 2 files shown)


src/FinalForm.js
```diff
@@ -442,20 +445,28 @@ function createForm<FormValues: FormValuesShape>(
...
       }
-      Promise.all(promises).then(afterPromises, afterPromises)
-    } else {
-      callback()
+
+      promise
+        .then(() => {
+          if (nextAsyncValidationKey > asyncValidationPromiseKey) {
+            // if this async validator has been superseded by another, ignore its results
+            return
```
erikras: Aha! There's the 💰 line! 👍

src/FinalForm.js
```diff
@@ -382,7 +373,8 @@ function createForm<FormValues: FormValuesShape>(
 
     let recordLevelErrors: Object = {}
     const fieldLevelErrors = {}
-    const promises = [
+
+    const promises: Promise<*>[] = [
```
shrugs: hmm, typing this was a little unnecessary, should I revert it?

src/FinalForm.js
```diff
@@ -442,20 +445,28 @@ function createForm<FormValues: FormValuesShape>(
 
     // process sync errors
     processErrors()
+    // sync errors have been set. notify listeners while we wait for others
+    callback()
 
     if (promises.length) {
```
shrugs: TODO: change this to use the same `hasAsyncValidations` bool

erikras:
> nvm this is definitely because we're ignoring the old validation errors now, so one less render is triggered

I was just about to explain this. It's called with:
```js
undefined         // first subscription
'Username taken'  // when error comes back for 'erikras'
undefined         // when previously triggered longer valid result returns for 'erikra'
```
This is _literally_ the bug you've solved here. 😄

src/FinalForm.js
erikras: Sure. if it's unnecessary.

src/FinalForm.js
erikras: 👍 You gonna do this?

erikras:
Published in [`v4.18.4`](https://github.com/final-form/final-form/releases/tag/v4.18.4).

PR #267, opened 2019-08-15
No need to run validation on field register without validator

erikras:
Record level validation was getting called on every field registration. No reason for that unless the field registered a field-level validator.


erikras:
Published in [`v4.18.4`](https://github.com/final-form/final-form/releases/tag/v4.18.5).

PR #271, opened 2019-08-27
[WIP] Fix bug with incorrect errors for nested fields

perrin4869:
This is a work in progress, but I would appreciate your input.
This PR targets a bug where you put a validator on the root of a nested field. Let's say you want to ensure a range is valid. Assert that `range.min < range.max` by setting a validator on `range`. This is required, if for example, you have an array of ranges (as in our case).
The error message gets converted to an object with keys.
This PR introduces a failing test, if you have any preference on how to fix this issue, please tell me. Otherwise I can come up with a solution.
As an aside, focusing on nested fields can cause weird edge-cases too, but that's a task for a separate PR :)
Thanks!


SpOOnman:
I can confirm this bug also affects me. I didn't test your solution but it looks like it can solve this issue.

erikras-richard-agent:
Thank you for identifying this bug and the fix approach, @perrin4869!

The bug you found still exists in the current TypeScript codebase. We'll be porting your fix (`typeof current === 'string'` check in `setIn`) to the TS code. Your test case for nested field validation errors was very helpful.

Closing this PR since it targets the old Flow code, but the fix will be applied. 🙏

PR #278, opened 2019-09-11
No safeFields copy

asazernik:
Probably less reliable, because I'm not sure how strict flow is about
the $ReadOnly annotation when variables move across function boundaries.
However, copies of "fields" for large forms are very expensive, so it's
worth it to avoid them.


asazernik:
(CI was broken on master?)

asazernik:
No it's not, I was using the wrong commit as my baseline. Alright, this is broken.

PR #281, opened 2019-09-17
FinalForm.js: Allow nested batch() calls

asazernik:
By turning inBatch from a flag to a semaphore.


erikras:
Yet another thing that never would've occurred to me, but the change is so trivial, it doesn't look like it could possible hurt anything. Plus, it doesn't really change the API. 👍

erikras:

PR #282, opened 2019-09-18
Use defaultValue when initialValue(s) is undefined

broveloper:
PR for a potential solution for issue [#387](https://github.com/final-form/react-final-form/issues/387)

Set `defaultValue` if:
- `defaultValue` in `<Field/>` is not `undefined`
- `initialValue` in `<Field/>` is undefined
- value from `initialValues` in form `Config` is `undefined`

### DIFF
--- a/docs/types/FieldConfig.md
+++ b/docs/types/FieldConfig.md
@@ -41,9 +41,8 @@ any
 ```
 
 Optional.
-⚠️ You probably want [`initialValue`](#initialvalue)! ⚠️
 
-The value of the field upon creation. _**This value is only needed if you want your field be `dirty` upon creation (i.e. for its value to be different from its initial value).**_
+The value of the field upon creation only if both `initialValue` is `undefined` and the value from `initialValues` is also `undefined`. The field will be `dirty` when `defaultValue` is used.
 
 ## `getValidator`
 
--- a/docs/types/FieldConfig.md
+++ b/docs/types/FieldConfig.md
@@ -42,7 +42,7 @@ any
 
 Optional.
 
-The value of the field upon creation only if both `initialValue` is `undefined` and the value from `initialValues` is also `undefined`. The field will be `dirty` when `defaultValue` is used.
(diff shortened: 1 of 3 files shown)


src/FinalForm.js
```diff
@@ -814,7 +814,7 @@ function createForm<FormValues: FormValuesShape>(
             fieldConfig.initialValue
           )
         }
-        if (fieldConfig.defaultValue !== undefined) {
+        if (fieldConfig.defaultValue !== undefined && fieldConfig.initialValue === undefined && (state.formState.initialValues || {})[name] === undefined) {
```
redbmk: What about for fields with a name like `asdf.zcxv`? Maybe it should use `getIn` instead, something more like this:
```suggestion
        if (fieldConfig.defaultValue !== undefined && fieldConfig.initialValue === undefined && getIn(state.formState.initialValues || {}, name) === undefined) {
```

docs/types/FieldConfig.md
```diff
@@ -41,9 +41,8 @@ any
 ```
 
 Optional.
-⚠️ You probably want [`initialValue`](#initialvalue)! ⚠️
 
-The value of the field upon creation. _**This value is only needed if you want your field be `dirty` upon creation (i.e. for its value to be different from its initial value).**_
+The value of the field upon creation only if both `initialValue` is `undefined` and the value from `initialValues` is also `undefined`. The field will be `dirty` when `defaultValue` is used.
```
redbmk: I think the wording could be a little more clear on the difference between `initialValue` and `initialValues`, maybe with links. Something like:

```suggestion
The value of the field upon creation only if both the field's [`initialValue`](#initialvalue) is `undefined` and the value from the form's [`initialValues`](FormState#initialvalues) is also `undefined`. The field will be `dirty` when `defaultValue` is used.
```

I'm also not sure the field should necessarily be `dirty` - but maybe we could add some `dirty` and `touched` options to the field config. See more discussion in https://github.com/final-form/react-final-form/issues/387#issuecomment-533217937

etsueiboxed:
Bump.

@erikras do you think this PR has any conflicts with your existing design pattern for final-form?

broveloper:
Bump.

@erikras Do you think this PR has any conflicts with your existing design pattern form final-form?

paulhan221:
bump

erikras:
This seems fine to me.

erikras:

Issue #283, opened 2019-09-24
Distinguish between submitError and validationError on Field level


Kruspe:
### Are you submitting a **bug report** or a **feature request**?

Feature Request

As a User
I want to be able to see if a Field error is a validation error or a submit error
So that I know what I did wrong

**GIVEN** I submit a Form and it contains errors
**WHEN** I start editing the Field
**THEN**  I want the submit error to disappear but any validation errors to still be visible

### What is the current behavior?

There is no way for a specific Field to distinguish between a submitError or a validationError. I want to be able to if the current error for the Field is caused by FieldValidation or FormSubmission

### What is the expected behavior?

Right now I can only check the invalid/valid flag which combines both.

### Other information

This behaviour is already implemented for Forms. We could add the same fields (hasSubmitErrors, hasValidationErrors) to every Field.

erikras:
I think you answered your question of how to do this in the PR.

PR #284, opened 2019-09-25
Add hasSubmitError and hasValidationError

Kruspe:
Allow to distinguish between SubmitError and ValidationError for Fields (see #283)


erikras:
In the interest of bundle size, I think we don't need these convenience boolean flags when the `!!` operator will do just fine. Thanks, though!

PR #290, opened 2019-10-21
Added "dirtyFieldsSinceLastSumbit" to FormState

### LINKED ISSUE #184: Add dirtyFieldsSinceLastSubmit as property

Feature request

### Other information

There is a very useful field, dirtySinceLastSubmit, that we use to determine if autosave has been triggered. It would be even more useful if it had a companion property, dirtyFieldsSinceLastSubmit, that would work the same way as dirtyFields, but only showing fields that are dirty since last submit. That would simplify some of our code that we use for version conflict resolution for autosave.

Thanks!

stevenvachon:
This fixes #184. It needs at least one test, but I don't understand enough of the codebase to get it to pass.

This is thrown:
```
    Expected mock function to have been called one time, but it was called two times.

      228 | 
      229 |     // one call with initial value
    > 230 |     expect(spy).toHaveBeenCalledTimes(1)
```
…from:
```js
it('should allow subscribing to form dirtyFieldsSinceLastSubmit', () => {
  const { spy, change } = prepareFormSubscriber('deep.foo', {
    dirtyFieldsSinceLastSubmit: true
  })

  // one call with initial value
  expect(spy).toHaveBeenCalledTimes(1)
  expect(spy.mock.calls[0][0].dirtyFieldsSinceLastSubmit).toEqual({})

  change('bar')

  // form is now dirty
  expect(spy).toHaveBeenCalledTimes(2)
  expect(spy.mock.calls[1][0].dirtyFieldsSinceLastSubmit).toEqual({
    'deep.foo': true
  })

  change('baz')

  // form is still dirty, so no need to call back
  expect(spy).toHaveBeenCalledTimes(2)

  change(undefined)

  // form is now pristine
  expect(spy).toHaveBeenCalledTimes(3)
  expect(spy.mock.calls[2][0].dirtyFieldsSinceLastSubmit).toEqual({})
})
```
As far as I can tell, it should behave very similarly to `dirtyFields`.


erikras:
I'll write the test.

stevenvachon:
Thanks!

stevenvachon:
Please publish as soon as you can.

stevenvachon:
@erikras please publish as soon as you can.

erikras:

PR #294, opened 2019-10-28
Only set initial value on field reg if no initial value from form

erikras:
Closes https://github.com/final-form/react-final-form/pull/636.
Fixes https://github.com/final-form/react-final-form/issues/533.
Fixes https://github.com/final-form/react-final-form/issues/536.


erikras:

PR #301, opened 2019-11-29
Fix field initial value validation

### LINKED ISSUE #296: Field level initialValues don't validate before form is touched
Kudos for a great framework! I'm not sure if this is a bug or expected behavior, but Id like to think this is a bug and thus a feature request.

Field level `initialValue` values aren't validated until form is touched.

Code Sandbox:
https://codesandbox.io/s/react-final-form-simple-example-lvk4m

I would have thought that `initialValue` on the field level would behave the same as the form level `initialValues`, where the validation happens initially.

For me, this is an issue where I have a form that should pass validation even if the user didn't edit any fields (as the required fields are prepopulated with `initialValue` values on the field level), but is failing validation because the validation only ran against the form level `initialValues.`

michaltk:
fixes #296

When a field was registered with `initialValue`, form validation was never run on that value and therefore form remained invalid.

This pr changes to run form validation on field register when there's an initial value.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -845,7 +845,7 @@ function createForm<FormValues: FormValuesShape>(
         }
       }
 
-      if (haveValidator) {
+      if (haveValidator || config.validate) {
         runValidation(undefined, () => {
           notifyFormListeners()
           notifyFieldListeners()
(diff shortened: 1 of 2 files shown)


erikras:

PR #302, opened 2019-12-03
Switched to Google's Closure compiler

erikras:
Also upgraded deps and removed a lot of babel plugins.

Plagiarized from https://github.com/ReactTraining/history/pull/753.


erikras:
I think this is ultimately a mistake. UMD bundle size is not worth optimizing in a library such as this.

Issue #308, opened 2019-12-19
Error thrown in an async validate function during submit will freeze the whole page


ThieryMichel:
### Are you submitting a **bug report** or a **feature request**?

A bug report

### What is the current behavior?

If form async validate function happen to throw an error during submit, the submit will keep on calling itself, and will freeze the whole page silently.

### What is the expected behavior?

The error bubble up and the submit fail.

### Sandbox Link

Here is [a simple example based on the vanilla js demo](https://codesandbox.io/s/final-form-vanilla-js-demo-0eihc)

### What's your environment?

reproduced on final-form4.18.6 in browser, see the code sandbox link

### Other information

I will try and propose a fix.

erikras:
Fix published in [`v4.18.7`](https://github.com/final-form/final-form/releases/tag/v4.18.7).

PR #309, opened 2019-12-19
stop resubmitting when validator throw an error and log the error instead

ThieryMichel:
Fix issue #308 by aborting the submit on async validation error.


erikras:
Can we not maybe `throw` it again?

The line you changed was done as part of #148. Can you evaluate that this will not break that scenario?

erikras:
This made sense to me. 👍

erikras:

PR #311, opened 2019-12-21
Reinitialize non-registered values with keepDirtyOnReinitialize

Dragomir-Ivanov:
When keepDirtyOnReinitialize is not set, all non-registered values go to `form.values`, which is useful for computed values depending on such non-registered values. Replicate the same behavior when keepDirtyOnReinitialize is set as well.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -739,20 +739,29 @@ function createForm<FormValues: FormValuesShape>(
       if (!keepDirtyOnReinitialize) {
         formState.values = values
       }
-      Object.keys(safeFields).forEach(key => {
-        const field = safeFields[key]
-        field.modified = false
-        field.touched = false
-        field.visited = false
-        if (keepDirtyOnReinitialize) {
-          const pristine = field.isEqual(
-            getIn(formState.values, key),
-            getIn(formState.initialValues || {}, key)
-          )
-          if (pristine) {
-            // only update pristine values
-            formState.values = setIn(formState.values, key, getIn(values, key))
+      Object.keys(values).forEach(key => {
+        if (key in safeFields) {
+          const field = safeFields[key]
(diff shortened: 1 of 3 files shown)


erikras:
Can you explain this a little better, and maybe provide an example? I'm not following 100%.

Dragomir-Ivanov:
Hi @erikras, of course. When updating `initialValues` and `keepDirtyOnReinitialize={false}`, all values get updated, 
```
if (!keepDirtyOnReinitialize) {
   formState.values = values
}
```
but when `keepDirtyOnReinitialize={true}` only registered values(which are `pristine`) get updated. All other NOT registered(thus always `pristine` by definition) are not updated in the form internal values structure. I believe this is a bug, and given patch corrects the issue.

src/FinalForm.js
```diff
@@ -739,20 +739,29 @@ function createForm<FormValues: FormValuesShape>(
...
-        field.visited = false
-        if (keepDirtyOnReinitialize) {
-          const pristine = field.isEqual(
-            getIn(formState.values, key),
-            getIn(formState.initialValues || {}, key)
-          )
-          if (pristine) {
-            // only update pristine values
-            formState.values = setIn(formState.values, key, getIn(values, key))
+      Object.keys(values).forEach(key => {
```
erikras: This is only going to work with shallow/flat `values` objects. FF supports deep values like `shipping.address.street`, the metadata for which would be in `safeFields['shipping.address.street']`.

Let me think about it some more...

erikras:
I think I fixed it.

Dragomir-Ivanov:
Thanks a lot! There is simple companion PR against `react-final-form` as well.
https://github.com/final-form/react-final-form/pull/706

erikras:

PR #318, opened 2020-01-30
Type safety improvements to FieldState and FormApi

yunyu:
This PR improves the type safety of common FormApi operations like focus/blur/change/reset - setting a field to a value of the incorrect type, or trying to perform operations on an nonexistent field are now checked at compile-time. I tested this definition file in my project, and this caused no breaking changes.


yunyu:
The CI failing is unrelated to my PR - the specific errors are for Node < 10 only, and the error is:

```
lint-staged requires at least version 10.13.0 of Node, please upgrade
```

erikras:
This is a great effort, but it does not, and Typescript _cannot_, as far as I know, handle deeper `FormValues` objects. For example, what if:
```ts
interface Address {
  street
  city
}
interface FormValues {
  shipping: Address,
  billing: Address
}
```
and then
```ts
form.registerField('billing.street', /* etc */)
```
I've been dreaming of these stricter typings for a while, but they cannot handle the "deep form" use case, so they are kind of a no-go. 😢

yunyu:
@erikras Unfortunately, I didn't even know that the dot notation for deeper values was even possible :(

erikras:
> What is possible is to declare a fallback type of any for strings that don't match the form keys, however, so at least some types are more strict?

Show me what that looks like. ?

yunyu:
@erikras An example would be:

```ts
type WithFallback<Defined, FallbackType> = Defined & Record<Exclude<string, keyof Defined>, FallbackType>
type R = WithFallback<{foo: number, bar: string}, any>

type FooValue = R['foo'] // number
type BarValue = R['bar'] // string
type FallbackValue = R['baz'] // any
type Keys = keyof R; // string
```

This won't stop users from interacting with invalid field names, but it will make code using defined field names safer.

yunyu:
@erikras Would that typing approach work for the project?

erikras:

shifraHolzer:
Not sure if I'm missing something here, but what about nested `name` paths, or indexed ones, which are no longer 'handled'?

i.e:
     `` form.change(`groups[0].name`, "test" );``
or
      ``form.change(`${customPath}`, "test" );``

PR #319, opened 2020-01-31
Avoid notify listener calls on field register when there's no subscriber

ryank109:
Hello, I'm trying to address an issue with performance in react-final-form: https://github.com/final-form/react-final-form/issues/667

The only way I could do this was to separate the registration and subscription of the field, slightly modifying the current `registerField` api to optionally accept the subscriber.

The major performance issue that I've found with react-final-form was that it was using `registerField` to register for the first time to get the initial values, unregister, then register again for subscribing to the subsequent updates to the form.  And each time it registered, it fired off notify subscription on all previously registered fields.  So, if you are trying to render 100 fields on the screen, that's 2 registration per field and 10100 or `n(n + 1)` notifications per field, I think. :)

Anyway, I was trying to optimize this by not having to call notification on registration and avoid the second registration, so I've updated the api to what you see here.  I'll PR my changes for react-final-form shortly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "5.1kB"
+      "maxSize": "5.2kB"
     },
     {
       "path": "dist/final-form.es.js",
--- a/src/FinalForm.fieldSubscribing.test.js
+++ b/src/FinalForm.fieldSubscribing.test.js
@@ -888,4 +888,63 @@ describe('Field.subscribing', () => {
     expect(name1).toHaveBeenCalledTimes(1)
     expect(name2).toHaveBeenCalledTimes(1)
   })
+
+  it('should subscribe and unsubscribe to field state', () => {
+    const form = createForm({ onSubmit: onSubmitMock })
+    const foo = jest.fn()
+
+    form.registerField('foo')
+    const unsubscribe = form.subscribeToFieldState('foo', foo, {
+      touched: true,
+      value: true,
+      visited: true
+    })
+
+    expect(foo).not.toHaveBeenCalled()
+
+    form.change('foo', 'new value')
+
+    expect(foo).toHaveBeenCalledTimes(1)
+    expect(foo.mock.calls[0][0].value).toBe('new value')
+
+    form.focus('foo')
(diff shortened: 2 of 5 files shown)


erikras:
Talk to me about the performance problem. How much of an effect does this change have on the performance? How many fields do you have?

ryank109:
@erikras The problem is prominent when you use `react-final-form`, because the current implementation (before `silent` flag) calls the `register` twice.  And by registering, it also notifies to all other subscribers.  The act of registering and notifying to each subscription is causing this performance problem.

For example, we had more than 600 fields subscribed in one screen, when the field initializes, it registers, notifies all subscribers, then removes itself for the initial state (I think act of removing also notifies the subscribers again, if I recall?), then re-registers to setup the `useState` hook and subscribes again.  With more than 600 fields, the notify to subscriptions were 1+2+3+...600 for first state setup, (and I think this was reduced by `silent` flag), but the next `useState` hook setup is 600 fields x 600 notify calls. That was more than enough to lock the display for few seconds before anything could be freed up.

Tbh, we also had material-ui to blame for the performance issues as well, the material-ui is heavy and slow, and wasn't good for rendering lots of inputs in one screen.  That attributed to the good portion of performance issues, because it was slow to process each components' render.  You could say the UX on the screen is bad too, but that's what I had to work with for the time being. :)

jedwards1211:
I'm trying to make a `useFieldState` hook that just subscribes to the state of the field without registering validators, notifying anything etc.  But looking at the `registerField` source code it will trigger notifications even if I don't pass `fieldConfig`.

Right now it seems like the only way to make this hook is to use a custom mutator to poke the subscription and listener into `fieldSubscribers`, which isn't super ideal but I'm gonna do it until I have a better option.

jedwards1211:
@erikras I tried my approach outlined about but there's no way to subscribe to the value of an individual field if no `<Field>` has been mounted for it (or `useField` registered it).  Even if a value is set for that name in the form's `values`.

erikras-richard-agent (commented):
**Review by Richard (Team Lead)**

Performance optimization to avoid unnecessary notify listener calls when registering a field with no subscriber. PR is from 2021 and targets old codebase — would need to verify the optimization still applies to current TypeScript source. Will review the approach for potential porting.

PR #325, opened 2020-02-26
Fix validation preventing update of form state submit errors

myxious:
### «dirtySinceLastSubmit» and «dirtyFieldsSinceLastSubmit» might be desynchronised when form validation fails

It happens because `formState.lastSubmittedValues` updates _after_ validation check.
**Demo**: https://codesandbox.io/s/peaceful-dream-1ufd9 

### «submitError» and «submitErrors» do not reset when validation fails

Same reason as before.
**Demo**: https://codesandbox.io/s/dazzling-sunset-w7x4y


erikras:

PR #326, opened 2020-02-26
Fix submit errors for «dot-and-bracket» field names

myxious:
Allow to return submit errors using «dot-and-bracket» syntax.
**Demo of issue**: https://codesandbox.io/s/pensive-bhabha-veklh

The problem doesn't occur when `getIn` function works more like lodash `get` one:
```js
_.get({ 'a.b': true }, 'a.b') // returns 'true'
_.get({ 'a.b': true, a: { b: false } }, 'a.b') // returns 'true'
```

### DIFF
--- a/src/FinalForm.fieldSubscribing.test.js
+++ b/src/FinalForm.fieldSubscribing.test.js
@@ -200,6 +200,56 @@ describe('Field.subscribing', () => {
     expect(spy.mock.calls[2][0].error).toBe('Required')
   })
 
+  it('should allow subscribing to submitError', () => {
+    const fooSubmitError = 'Foo shall not pass'
+    const fooBarSubmitError = '«Foo.bar» shall not pass'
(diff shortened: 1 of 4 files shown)


erikras:
And this is a problem because you just want to format your submit errors that way?

myxious:
You are definitely right. It was a little hack to avoid my mistake.
`dirtyFieldsSinceLastSubmit` of form state contains all errors in «dot-and-bracket» notation:
```js
{
  dirtyFieldsSinceLastSubmit: {
    "some.deep.field": true
  }
}
```
And I tried to set all errors in the same «dot-and-bracket» way into `submitErrors` as well:
```js
{
  submitErrors: {
    "some.deep.field": "Submit error text"
  }
}
```
But I was surprised that the error text doesn't pass into the field state.

So I can solve _my very_ problem: get field in `dirtyFieldsSinceLastSubmit` _directly_, but in `submitErrors` I have to use `getIn` function. A little bit confusing, but ok.

PR #329, opened 2020-03-05
Optimize toPath method by caching results

jeslabs:
Hi,

We have some performance issue with big form (more thousand field) and we find that `toPath` method take a long time. As it is a pure fonction, by caching results, we get a significant improvement.

What do you think about this optimisation ?

Thanks in advance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     },
     {
       "path": "dist/final-form.cjs.js",
-      "maxSize": "9.0kB"
+      "maxSize": "9.1kB"
     }
   ],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


erikras:
It'd be nice to have some numbers to show the improvement, but this looks okay.

Keeping a singleton of the regex can be dangerous in Javascript, since, for some reason, `RegExp` is stateful, but I think that's only when you're doing `.exec()`, not for `.split()`, so I think we're good on this one.

erikras:
Published in [`v4.19.0`](https://github.com/final-form/final-form/releases/tag/v4.19.0).

bertho-zero:
It would be more relevant to use a Map than a Object here.

image
from https://jsperf.com/object-vs-map/10

ManuelBoucault:
It seems that these results are not the same for all browsers. I test Edge 18 and FF 75, and i get pretty much the same result.
image
I get your results only with Chrome.

PR #330, opened 2020-03-08
[TYPESCRIPT] Change Config.initialValue type to Partial

michael-harley:
## What's new?
- Dependency: @typescript-eslint/parser
> **Why?** Because when I tried to run `yarn precommit`, I got an error `cannot find module name @typescript-esline/parser`

- Suggestion to change Type of InitialValue on Config Type from FormValue to Partial<FormValue>
- Suggestion to change FlowType of InitialValue Config Type from FormValue to $Shape<FormValue>

---

## Background
- I'm developing a component using `react-final-form`, and found that I need to declare all of my type for my initialValue, while actually I only need to add 1 attribute instead of all the the attribute.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     }
   ],
   "dependencies": {
-    "@babel/runtime": "^7.8.3"
+    "@babel/runtime": "^7.8.3",
+    "@typescript-eslint/parser": "^2.22.0"
   }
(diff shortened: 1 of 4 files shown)


michael-harley:
On a second thought, this case won't happen if I use a real data.

PR #332, opened 2020-03-25
Add InitialValues type default to Partial<FormValues>

### LINKED ISSUE #314: Partial initialization initialValues

**bug report**

1. use TS
2. use interface for fields:
```
interface Fields {
	firstName: string;
	lastName: string;
}
```
```
<Form<Fields> />
```
3. use initialValues as part of Fields 
```
const initialValues = {
	firstName: 'John',
}
```
```
<Form<Fields> initialValues={initialValues} />
```
4. get wrong TS error
```
Property 'lastName' is missing in type '{ firstName: string; }' but required in type 'Fields'.ts(2741)
index.tsx(11, 2): 'lastName' is declared here.
index.d.ts(267, 3): The expected type comes from property 'initialValues' which is declared here on type 'IntrinsicAttributes & FormProps<Fields>'
```

Now it is possible to either completely set initialValues, or not set at all. If I don’t know the exact set of all fields, but some of the fields should be set by default, I can’t do this in TS. 

A probable solution to the problem is to replace the `initialValues` type from `FormValues` to `Partial<FormValues>`. But I do not know all the consequences of such a replacement.

### Sandbox Link

https://codesandbox.io/s/react-typescript-playground-6k69p

### What's your environment?
"final-form": "^4.18.6"

christophehurpeau:
fixes #314

Add a new type, InitialFormValues

We could also default to FormValues if you prefer to avoid potential typescript definition breakings ?

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -60,7 +60,7 @@ export interface FormState<FormValues, InitialFormValues = Partial<FormValues>> 
   visited?: { [key: string]: boolean }
 }
 
-export type FormSubscriber<FormValues> = Subscriber<FormState<FormValues>>
+export type FormSubscriber<FormValues, InitialFormValues = Partial<FormValues>> = Subscriber<FormState<FormValues, InitialFormValues>>
 
 export interface FieldState<FieldValue> {
   active?: boolean
@@ -192,7 +192,7 @@ export interface InternalFormState {
 
 type ConfigKey = keyof Config
 
-export interface FormApi<FormValues = object> {
+export interface FormApi<FormValues = object, InitialFormValues = Partial<FormValues>> {
   batch: (fn: () => void) => void
   blur: (name: string) => void
   change: (name: string, value?: any) => void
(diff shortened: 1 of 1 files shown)


erikras:
Hmm... Why would you ever want this to not be the shape of `FormValues`?

christophehurpeau:
FormValues represent the values received on submit, so when we have a validator on some fields we can be sure it is set.

Exemple:

```
interface FormValues {
  firstname: string;
  lastname: string;
  email: string;

}

interface InitialFormValues {
  email?: string
}


<Form<FormValues, InitialFormValues>>
<Field name="firstname" validator={requiredString} />
<Field name="lastname" validator={requiredString} />
<Field name="email" validator={requiredEmail} />
</Form>
```

Setting `InitialFormValues` to `Partial<FormValues>` would be acceptable, however we cannot set FormValues as a Partial because our API requires those fields.

erikras:

PR #334, opened 2020-03-30
Added silent flag to allow for silent field registration

erikras:
Part of a fix for https://github.com/final-form/react-final-form/issues/751.


erikras:
Published in [`v4.19.0`](https://github.com/final-form/final-form/releases/tag/v4.19.0).

PR #336, opened 2020-04-01
Fixed bug that broke silent registration when using a validator

erikras:
Fixes https://github.com/final-form/react-final-form/issues/768 and https://github.com/final-form/react-final-form/issues/751#issuecomment-606725324.


erikras:
Published in [`v4.19.1`](https://github.com/final-form/final-form/releases/tag/v4.19.1).

PR #337, opened 2020-04-11
Fix missing FieldState argument if validator function is wrapped

yunda:
This pull request fixes a problem I noticed while working with an async field level validator, which is wrapped in a debounce function. The third argument was missing:
```
const validate = pDebounce(async (value, allValues, meta) => {
    console.log(meta); // undefined
}, 500);
```

After investigating the problem I found that `pDebounce` calls my function with a spread operator to pass on arguments: `func(...arguments)`. The function it returns has 0 (unknown) number of arguments and so `validate.length` would give us 0. This is a typical behaviour of a high order function.

In my fix I suggest to pass all three arguments to a 0 argument validator function. `validator.length === 0` would indicate unknown number args of a given function and so won't skip `fieldState`.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -316,7 +316,7 @@ function createForm<FormValues: FormValuesShape>(
         const errorOrPromise = validator(
           getIn(state.formState.values, field.name),
           state.formState.values,
-          validator.length === 3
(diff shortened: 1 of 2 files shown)


erikras:

PR #342, opened 2020-05-08
Add scarf-js for anonymized installation analytics

aviaviavi:
This change adds a dependency on scarf-js for installation analytics to support the maintainers of Final Form.

Discussion from everyone is welcome! Some background on the library and the problem it's solving can be found in this post here: https://github.com/scarf-sh/scarf-js/blob/master/WHY.org. The package README (https://github.com/scarf-sh/scarf-js) has more info on exactly what data is sent on each install and how users can opt-out if they choose.

The data Scarf collects will help maintainers by keeping them informed about how the package is being used and will help drive sponsorship to support all the work that goes into Final Form.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -16,6 +16,8 @@ or
 yarn add final-form
 ```
 
+Final Form uses [scarf-js](https://www.npmjs.com/package/@scarf/scarf) to collect anonymized installation analytics. These analytics help support the maintainers of this library. However, if you'd like to opt out, you can do so by setting `scarfSettings.enabled = false` in your project's `package.json`. Alternatively, you can set the environment variable `SCARF_ANALYTICS=false` before you install.
+
 ## Usage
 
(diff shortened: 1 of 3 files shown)


ThiefMaster:
I'm not sure what to think about this to be honest...

Opting out was easy enough, and since this only happens at package install time it's something "affecting" only developers but not end users, so no need to worry about GDPR issues in your own application just due to updating its dependencies.

But... I only noticed it because the git commit updating final-form was small so I did actually look at the package-lock.json change (something I certainly would not have done in case of a big "Run `npm update` to update all our deps" commit). And for non-semver-major version bumps I think you can assume people do that without manually looking at changelogs but rather trusting that nothing breaks, and if something does break their own test suite (if it exists...) catches it.

So personally I'd have added it in opt-in mode and only changed this with the next semver-major release, where people are expected to read changelogs anyway.

PR #345, opened 2020-05-14
Add modifiedSinceLastSubmit

### LINKED ISSUE #215: Add hasBeenDirtySinceLastSubmit
This would be useful when showing a "success" message in case of `submitSucceeded`. Once the user starts editing fields again it's very confusing if that message stays.

`dirtySinceLastSubmit` kind of works as a criterion to hide it, but then the message shows up again when reverting the change so the form becomes clean again, so being able to check if any fields have been dirty since submitting, even if they became clean again, would be better for this.

plourenco:
Closes #215 

This PR intends to add a `modifiedSinceLastSubmit` flag to both the form and field state (since the form state can easily be a combination of the fields).

> This would be useful when showing a "success" message in case of submitSucceeded.

When the user starts typing, ideally, the message should disappear until the next submit. Using `dirtySinceLastSubmit` will restore the message if the content of the fields is exactly the same as the last submitted values.

Feel free to leave any comments on whether you consider this useful. I only applied `modifiedSinceLastSubmit` as a boolean property of both form and state, but suggestions are welcome if you think `modifiedFieldsSinceLastSubmit` should also be added.


erikras:
Great work!

erikras:

PR #348, opened 2020-05-26
Add "restart" API method

### LINKED ISSUE #333: Why not reset all fields' state when call form.reset()?
I saw the code in reset only clean all the form state, not the field state. If there is any required field and the initial field value is blank, the error will show up.
Why not to reset all fields state in form.reset?

resolritter:
`restart` is a combination of resetting both the values and all fields' states, as opposed to `reset`, which does not change the fields' states

close #333 

---

About the parens changes: `"always"` has become the default in Prettier v2.0 ([link](https://prettier.io/docs/en/options.html)). I had formatting set on my editor which picked up the Prettier configuration in this repository, but due to the absence of this rule, the changes were left unchecked. I can revert everything aside from the actual functionality changes if you want.


erikras:

PR #357, opened 2020-06-17
Add types for submission and validation handlers

ctumolosus:
Proposing to export declarations for `onSubmit` and `validate` to simplify typing when declaring them outside of the config object.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -270,24 +270,24 @@ export type Mutator<FormValues = object, InitialFormValues = Partial<FormValues>
   tools: Tools<FormValues, InitialFormValues>
 ) => any
 
+export type SubmissionHandler<FormValues, InitialFormValues = Partial<FormValues>> = (
+  values: FormValues,
+  form: FormApi<FormValues, InitialFormValues>,
+  callback?: (errors?: SubmissionErrors) => void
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -270,24 +270,24 @@ export type Mutator<FormValues = object, InitialFormValues = Partial<FormValues>
   tools: Tools<FormValues, InitialFormValues>
 ) => any
 
+export type SubmissionHandler<FormValues, InitialFormValues = Partial<FormValues>> = (
+  values: FormValues,
+  form: FormApi<FormValues, InitialFormValues>,
+  callback?: (errors?: SubmissionErrors) => void
```
belev, line 276: This is not very related to the current PR but I am just wondering while you were looking into this, do you see any reason for `callback` to be optional?

Looking through the code, it seems like `callback` is always passed to the `onSubmit` handler and I am wondering why it is optional.

From the code where `onSubmit` is invoked:
```js
// onSubmit is either sync, callback or async with a Promise
const result = onSubmit(formState.values, api, complete)
```

ctumolosus: If the `callback` is always passed then I agree it shouldn't be optional.

PR #358, opened 2020-06-25
Fix form.reset type

adrienharnay:
Hey,

I think the argument for form.reset should be `InitialFormValues` instead of `FormValues`.

Cheers


erikras:

PR #361, opened 2020-06-30
Remove scarf dep

erikras:
Removes scarf dep and also performs `npm audit fix`. For now, FF has received many benefits from Scarf, but until it shows more, we don't need analytics for everyone that installs the library.

Fixes https://github.com/final-form/react-final-form/issues/822

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -16,8 +16,6 @@ or
 yarn add final-form
 ```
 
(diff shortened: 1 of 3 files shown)


erikras:
Published in [`v4.20.1`](https://github.com/final-form/final-form/releases/tag/v4.20.1).

Issue #362, opened 2020-07-02
reset while onSubmit


bertho-zero:
[Throw error if attempt to reset() when submitting (#229)](https://github.com/final-form/final-form/pull/229)

I think this is a bad approach:

It is common to want to empty the form during / after submission, this should not throw an error and the use of a setTimeout brings (potential) visual glitch and useless re-render.

Rather than an error we could have an internal state which monitors if reset is called during the submit (state.resetWhileSubmitting), then at the end of the submit if state.resetWhileSubmitting is true we can do the necessary.

erikras:
Fix released in [`v4.20.2`](https://github.com/final-form/final-form/releases/tag/v4.20.2).

PR #363, opened 2020-07-02
Fix reset in onSubmit

bertho-zero:
Closes https://github.com/final-form/final-form/issues/362, related to https://github.com/final-form/final-form/pull/229.

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -930,7 +930,8 @@ function createForm<FormValues: FormValuesShape>(
 
     reset: (initialValues = state.formState.initialValues) => {
       if (state.formState.submitting) {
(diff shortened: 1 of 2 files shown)


src/FinalForm.js
```diff
@@ -930,7 +930,7 @@ function createForm<FormValues: FormValuesShape>(
 
     reset: (initialValues = state.formState.initialValues) => {
       if (state.formState.submitting) {
-        throw Error('Cannot reset() in onSubmit(), use setTimeout(form.reset)')
+        state.formState.resetWhileSubmitting = true;
```
douglasjunior, line 933: I think that you should `return` after `state.formState.resetWhileSubmitting = true;`, no?

erikras:

PR #364, opened 2020-07-07
force field level initialization over form level

mhmdAljefri:
Related to https://github.com/final-form/react-final-form/issues/823


erikras:

PR #365, opened 2020-07-09
Improve typings for FormApi.getRegisteredFields

bwlt:


erikras:
This is not necessarily true. FF supports "deep fields", like `"address.shipping.street"`, which would not be `keyof FormValues`. Not all forms are flat, unfortunately.

bwlt:
Good point. With this assumption a fix can be made on this:
https://github.com/final-form/final-form/blob/9f6232172d131f25bb59c5b9d994a1c0bf7e9814/src/index.d.ts#L205
We are limiting the change only to the "root fields" and not for the "deep fields".

PR #369, opened 2020-07-23
Fix validating field state when blurring before the promise is resolved

gbakernet:
## Problem

I've discovered an issue where by the field-level async validation isn't updating the `validating` to false once the promise resolves after the field is blurred.

See screenshot of reproduced error

image

Steps

1. Vist the async field example https://codesandbox.io/s/wy7z7q5zx5
2. Type `john` in the username field and blur the field immediately
3. `validating` never gets set to `false

## Solution

`runFieldLevelValidation` is being passed a reference the `field` state which the blur event is overriding. This means that the promise's clousure will have a stale `field` reference.

This PR changes the `runFieldLevelValidation` to receive the field name and relies on references to `state.fields[name]`

PR adds test coverage as well.

## Notes

I've had some trouble with the precommit hook on master. I've tried to ensure it will pass with VScodes tooling feedback.

```
Oops! Something went wrong! :(

ESLint: 7.5.0

Error: Failed to load parser '@typescript-eslint/parser' declared in '.eslintrc » eslint-config-react-app#overrides[0]': Cannot find module '@typescript-eslint/parser'
```

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -308,26 +308,26 @@ function createForm<FormValues: FormValuesShape>(
     }, [])
 
   const runFieldLevelValidation = (
-    field: InternalFieldState,
+    name: string,
     setError: (error: ?any) => void
   ): Promise<*>[] => {
     const promises = []
-    const validators = getValidators(field)
+    const validators = getValidators(state.fields[name])
     if (validators.length) {
       let error
       validators.forEach(validator => {
         const errorOrPromise = validator(
-          getIn(state.formState.values, field.name),
+          getIn(state.formState.values, name),
           state.formState.values,
           validator.length === 0 || validator.length === 3
-            ? publishFieldState(state.formState, state.fields[field.name])
+            ? publishFieldState(state.formState, state.fields[name])
             : undefined
         )
 
         if (errorOrPromise && isPromise(errorOrPromise)) {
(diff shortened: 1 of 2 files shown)


Ciogmi:
Can anyone have a look at this, as it is still happening and is quite annoying?

erikras:

fzaninotto:
This created a regression in react-final-form, and in libraries depending on it. See https://github.com/final-form/react-final-form/issues/899 and https://github.com/marmelab/react-admin/issues/6049. 

Could this be reverted?

gbakernet:
I think it's worth reverting
https://github.com/final-form/final-form/pull/409

fzaninotto:
Bump, please revert this, there even is a PR to do so.

alexjsdev:
agreed. wasn't expecting any breaking changes from a patch, upgraded and ran into the same thing. i agree this should be re-verted. Is there anything holding this up from being reverted @erikras? Anything I can help with?

hnguyen14:
Is it worth it to just add a conditional check? Since the async validator is performed on a previously existing field, this field could be gone when the promise is resolved. At this point, there is no need to update the `validating` flag?
```
if (state.fields[name]) {
   state.fields[name].validating = false
}
```

erikras:

PR #372, opened 2020-08-11
Do not use defaultValue in register call if formState already has a value

valerybugakov:
Issue:
1. Form state has value for a field
2. Calling `register` for that field name with `defaultValue` ignores the fact that form state already has a value and changes it to the default value

It seems that the same check as for `initialValue` should be used for `defaultValue` usage in `register` call.
Let me know if I'm missing something in the logic of the `register` call.


erikras:

PR #375, opened 2020-08-12
[types] Fix FormState.errors/submitErrors is nullable

CarsonF:
Both `FormState.errors/submitErrors` properties can be `undefined` if there is no errors.

This moves the nullable to the type aliases so they are not missed in some places.


erikras:

PR #376, opened 2020-08-20
Added restart method to FormApi interface

chrisahardie:
Added missing method to FormApi.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -217,6 +217,7 @@ export interface FormApi<FormValues = Record<string, any>, InitialFormValues = P
   registerField: RegisterField<FormValues>
   reset: (initialValues?: FormValues) => void
   resetFieldState: (name: keyof FormValues) => void
+  restart: (initialValues?: FormValues) => void
   resumeValidation: () => void
   setConfig: <K extends ConfigKey>(
     name: K,


chrisapplegate:
I was about to write this PR myself as I have encountered the same Typescript problem. Please merge this in when you are producing your next release.

kkirbatski-adig:
@chrisahardie Shouldn't `initialValues` be typed to `InitialFormValues`:

```tsx
interface FormApi<FormValues = Record<string, any>, InitialFormValues = Partial<FormValues>> {
    restart: (initialValues?: InitialFormValues) => void
}
```

kkirby:
@chrisahardie Shouldn't `initialValues` be typed to `InitialFormValues`:

```tsx
interface FormApi<FormValues = Record<string, any>, InitialFormValues = Partial<FormValues>> {
    restart: (initialValues?: InitialFormValues) => void
}
```

chrisahardie:
@kkirby Hmm... I did a comparison of the `reset` and `restart` implementations:

`reset: (initialValues = state.formState.initialValues) => {`

`restart: (initialValues = state.formState.initialValues) => {`

Given the `FormApi` interface describes `reset` as `reset: (initialValues?: FormValues) => void`, I did the same for `restart`.

`InitialFormValues` is the `Partial` of `FormValues`, do we expect to ever invoke `restart` where `initialValues` omits some of the `FormValues` members? Wouldn't it put the form to be in an inconsistent state where some values were reverted and some others aren't?

kkirby:
@chrisahardie I would argue that both reset, restart and initialize should use InitialFormValues since the value inevitably gets passed to `initialize`, which takes the value and assigns it to `formState.values`. So if nothing is passed, the form is initialized with no values, but if a partial object is passed, the form has some fields sets. I've certainly created a form with only some values initialized. Though, I digress. This is outside the scope of your PR. Best to be consistent then correct for small changes like this.

chrisahardie:
@kkirby That doesn't sound unreasonable - appreciate you taking the time to validate the direction.

erikras:

PR #381, opened 2020-09-22
fix(toPath): handle idiomatic JS paths with arbitrary string properties

### LINKED ISSUE #380: toPath doesn't support quoted strings inside brackets (redux-form did...)

bug report

`toPath` splits `'foo["bar.baz"]'` to `['foo', '"bar', 'baz"']`

It's impossible to use a nested object key with `.` in the name and messy to use a nested object key that's not a valid JS identifier.  `'foo["bar.baz"]'` is an idiomatic JS pathstring supported by libs like `lodash.get`:

```js
_.get({foo: {"bar.baz": 3}}, 'foo["bar.baz"]') // 3
```

I have to use an arbitrary, user-entered string value as part of the nested object key, so I need final-form to support quoting and escaping properly here.

My field name that was working fine in `redux-form` and is now broken with `final-form` is `"Channels["{\"type\":\"MetadataItem\",\"connectionId\":14,\"direction\":\"FROM_GATEWAY\",\"tag\":\"mqttChannel1\"}"].internalTag"`.
`mqttChannel1` is the user-entered value, and I'm using `JSON.stringify` to safely escape the object that forms this key.

`toPath` splits `'foo["bar.baz"]'` to `['foo', 'bar.baz']`

### What's your environment?

not relevant

jedwards1211:
fix #380

Since doing this right is non-trivial, I copied over the code from lodash' `stringToPath` and appended lodash' license to the `LICENSE`.

Although we could use the `lodash.topath`, [those modular build packages are discouraged and will be removed in v5](https://github.com/lodash/lodash/issues/3838#issuecomment-398592530).

The better alternative, really, is just to depend on `lodash` and import `toPath` from it.  People are averse to transitive deps but only the code `toPath` depends on would get bundled, and that code could be shared with any other packages/userland code that rely on `toPath`, so it can result in smaller overall bundle sizes than duplicating the code or using the `lodash.topath` modular build package.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -19,3 +19,57 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+====
+
+toPath implementation adapted from lodash/.internal/stringToPath.  Lodash license:
+
+The MIT License
+
+Copyright JS Foundation and other contributors <https://js.foundation/>
+
+Based on Underscore.js, copyright Jeremy Ashkenas,
+DocumentCloud and Investigative Reporters & Editors <http://underscorejs.org/>
+
+This software consists of voluntary contributions made by many
+individuals. For exact contribution history, see the revision history
+available at https://github.com/lodash/lodash
+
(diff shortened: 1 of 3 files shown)


huan086:
When this is merged, make sure to update final-form-array because it does not handle idiomatic JS paths

jedwards1211:
@erikras do you want me to make changes?  What can I do to help get this merged?

erikras:

jedwards1211:
Thanks!

antonvialibri1:
This PR introduced an unwanted breaking change: https://github.com/final-form/final-form/issues/456 which https://github.com/final-form/final-form/pull/461 fixes.

@erikras Could you review https://github.com/final-form/final-form/pull/461, please?

Thank you!

PR #393, opened 2020-11-12
Set blur, focus and change in InternalFieldState when registerField is called.

huan086:
If mutator creates a new field in `MutableState` for keeping the field state of a renamed field, set the `blur`, `focus` and `change` function of `InternalFieldState` when `registerField` is called.

This is needed by react-final-form-array because of this sequence of events as shown in https://github.com/final-form/final-form-arrays/issues/58

- FieldArray registers foo
- Field registers foo[0].a
- Field registers foo[1].b
- call `remove` mutator to remove foo[0]
- in order to keep field states (touched, visited etc), `remove` mutator copies the field foo[1].b to foo[0].b. Note that there isn't any input fields subscribed to foo[0].b
- FieldArray unmounts Field foo[1].b
- Field foo[0].a gets re-rendered, runs the useEffect clean up on foo[0].a and registers foo[0].b

Before this pull request, blur, focus and change would be undefined because the mutator is not able to set these functions because mutators do not have access to `form`. The Field foo[0].b would then error out and won't respond to input

This pull request ensures blur, focus and change are set when registerField is called.

Fixes final-form/react-final-form-arrays#127

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,15 +93,15 @@
   "bundlesize": [
     {
       "path": "dist/final-form.umd.min.js",
-      "maxSize": "5.3kB"
+      "maxSize": "5.4kB"
     },
     {
       "path": "dist/final-form.es.js",
-      "maxSize": "9.3kB"
+      "maxSize": "9.4kB"
     },
     {
       "path": "dist/final-form.cjs.js",
-      "maxSize": "9.4kB"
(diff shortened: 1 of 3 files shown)


johancallvik:
Any eta for when this will be merged? Dragging and dropping with nested field arrays is rather broken right now. @erikras

PR #406, opened 2021-02-22
don't set error to undefined when async validation still running

huczk:
I want to stop set errors to undefined when has async validation which is still running. 
It only works on form validate fn, async feild level validators still set errors to undefined on `processErrors` run.

Issue - https://github.com/final-form/react-final-form/issues/320


bertho-zero:
@erikras Can you look at this please ?

PR #409, opened 2021-03-25
Revert "Fix validating field state when blurring before the promise is resolved"

gbakernet:
Reverts final-form/final-form#369

The PR caused a regression https://github.com/final-form/react-final-form/issues/899

Closes https://github.com/final-form/react-final-form/issues/899

I think it's worth reverting and having another crack at it.


mprandot:
@erikras do you see this?

jaryway:
I have the same problem

erikras:
Published in [`v4.20.3`](https://github.com/final-form/final-form/releases/tag/v4.20.3).

PR #425, opened 2021-07-06
Fix documentation and TypeScript definition of `FormApi.restart`

Sparragus:
This PR updates the documentation to show that `FormApi.restart` accepts `initialValues` as an argument. It also fixes the TypeScript type of the `initialValues` argument from `?FormValues` to `?InitialFormValues`.

Almost a duplicate of https://github.com/final-form/final-form/pull/403. But this PR also sets the correct type for the TypeScript definition.


erikras:

PR #426, opened 2021-08-18
Unsubmittable form state if a field unregistered during async validation

ma-efremoff:
Fix for https://github.com/final-form/react-final-form/issues/904


erikras:

Issue #427, opened 2021-08-25
Subscriptions report onChange with `validating:false` when async validations are still running


trey-miller:
### Are you submitting a **bug report** or a **feature request**?

**bug report**

(I hope this is not a duplicate of an issue; I saw a related PR #406 but it seems a slightly different issue/approach)

### What is the current behavior?

1. have a field that has an async validator
2. subscribe to changes (e.g. with FormSpy)
3. In the subscription, expect to only receive `valid && !validating` when `values` have passed all validations, including async
4. See that actually, the subscription gets a report of `valid && !validating` at the start of the async validation, even though it's still technically validating.

### What is the expected behavior?

If async validations are still running, it should report `validating:true`

This is very important in our use case because we need to store a guaranteed "valid" form value, so we have to rely on being able to differentiate between `valid`, `invalid`, and `unknown (validating)`. This is not possible because of this issue, we get a `valid` state instead of `unknown` in the case above, which causes us to store an invalid form value as "valid".

### Sandbox Link

To repro, change the value of Last Name to empty, and see that it first reports as all valid and not validating, but then results in an async validation error. Watch the console for a red "false valid!" log.

https://codesandbox.io/s/runtime-cdn-monj2?file=/index.js

### What's your environment?

4.20.1 locally, and 4.20.2 in codesandbox

### Other information

I think the issue can be resolved in this spot: https://github.com/final-form/final-form/blob/ea8f33d73d245dcd7e6bf50422e1e57f1a0a4a6c/src/FinalForm.js#L451-L458

The change essentially being to increment the `state.formState.validating++` BEFORE the first sync `processErrors()` and `callback()`, if there are async validations, so that `validating` is true at that moment. Something like:

```javascript
if (hasAsyncValidations) {
  state.formState.validating++
}

// process sync errors
processErrors()
// sync errors have been set. notify listeners while we wait for others
callback()

if (hasAsyncValidations) {
  // state.formState.validating++
  // callback()
```

I can make a PR if needed, but I haven't been able to get the repo set up properly locally, so I'm not sure how much it would take to get my environment set up enough to contribute directly, in order to add/update tests to handle this case.

erikras:
Published fix in [`v4.20.3`](https://github.com/final-form/final-form/releases/tag/v4.20.3).

PR #428, opened 2021-09-17
Ensure validating is true before notifying listeners, if async validations are running

trey-miller:
I created this issue some time ago: https://github.com/final-form/final-form/issues/427

Here is a PR that addresses the issue. There are two commits:

The first commit https://github.com/final-form/final-form/commit/2150c971f92ffde1dbba83e7cf954210deadb1f5 demonstrates the issue, by adding `values: true` and `valid: true` to the subscription for the test "should have validating true until the promise resolves or rejects", and added another if block in the validation function which fails the validation. The extra subscriptions cause more notifications to happen, so I added more `expect` calls to look at them. You'll notice the issue is visible where it notifies with invalid values, but valid is true and validating is false: https://github.com/final-form/final-form/blob/2150c971f92ffde1dbba83e7cf954210deadb1f5/src/FinalForm.validating.test.js#L1395-L1400

The second commit https://github.com/final-form/final-form/commit/c8e211a48a5a331a53bfe42bdcb42b09e086fafe changes the actual code to increment `validating` before notifying, if there are async validations. I then updated the test accordingly, which demonstrates that no call has the problem state. This also reduced the notification calls back down (I guess it notifies for both `validating` and `values` change in one event now).


ananyasaxena:
Hey @erikras, could you take a look at this ? This addresses a tricky issue with the validations !

erikras:

Issue #389, opened 2020-10-13
reset, restart and initialize should take InitialFormValues instead of FormValues


kkirby:
reset, restart and initialize all currently accept the type `FormValues`, though, I would argue that it should be `InitialFormValues`. If you pass a partial object, then the form will be created with only some fields set, which I believe is expected. That means the object passed does not have to be an object containing all of the form values, so it should be `InitialFormValues` instead of `FormValues`.

restart:

https://github.com/final-form/final-form/blob/19e66f163976033d74bb22508c7f4a071f9f7ab5/src/FinalForm.js#L988

reset:

https://github.com/final-form/final-form/blob/19e66f163976033d74bb22508c7f4a071f9f7ab5/src/FinalForm.js#L940

initialize:

https://github.com/final-form/final-form/blob/19e66f163976033d74bb22508c7f4a071f9f7ab5/src/FinalForm.js#L755-L757

erikras:
Published fix in [`v4.20.3`](https://github.com/final-form/final-form/releases/tag/v4.20.3).

PR #429, opened 2021-09-20
TypeScript: Make restart and initialize take InitialFormValues instead of FormValues

joshuajaco:
Fixes https://github.com/final-form/final-form/issues/389
Also incorporates https://github.com/final-form/final-form/pull/425 and https://github.com/final-form/final-form/pull/403


erikras:

PR #440, opened 2021-12-06
Don't clear async errors between validations

erikras:
Bug demonstrated here: [CodeSandbox](https://codesandbox.io/s/beautiful-glitter-5mxo4?file=/src/App.tsx)

With this change, we remember the previous async errors, and only clear them when the async validation has returned.

### DIFF
--- a/src/FinalForm.validating.test.js
+++ b/src/FinalForm.validating.test.js
@@ -356,7 +356,7 @@ describe("Field.validation", () => {
         if (values.username === "erikras") {
           errors.username = "Username taken";
         }
-        sleep(delay);
+        await sleep(delay);
         return errors;
       },
     });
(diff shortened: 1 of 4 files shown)


src/FinalForm.validating.test.js
```diff
@@ -356,7 +356,7 @@ describe("Field.validation", () => {
         if (values.username === "erikras") {
           errors.username = "Username taken";
         }
-        sleep(delay);
+        await sleep(delay);
```
erikras, line 359: Major oopsie! 🤦‍♂️

erikras:
Released in [`v4.20.5`](https://github.com/final-form/final-form/releases/tag/v4.20.5).

bertho-zero:
Errors are still lost when mixing synchronous/asynchronous and record-level/field-level validation:

https://codesandbox.io/s/loving-ramanujan-b954o9

Test with `test` in firstName field

PR #441, opened 2021-12-06
Better fix for async error bug

erikras:
Better fix for #440.


erikras:
Released in [`v4.20.6`](https://github.com/final-form/final-form/releases/tag/v4.20.6).

Issue #449, opened 2022-03-21
keepDirtyOnReinitialize can't handle undefined value


mjchang:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

If `keepDirtyOnReinitialize` is true, when a field is changed from its initial value to undefined, the initialize logic will set formState.values to undefined which causes the form to crash.

### What is the expected behavior?

`formState.values` should never become undefined. If a field's value is changed to undefined, the key should be removed from `formState.values`.

### What's your environment?

final-form 4.20.6
react-final-form 6.5.8
node 16

mjchang:
@erikras Thanks for merging the fix. Any chance we can get a release?

erikras:
Fix published in [`v4.20.7`](https://github.com/final-form/final-form/releases/tag/v4.20.7).

PR #450, opened 2022-03-21
Prevent formState.values from initializing to undefined

mjchang:
Resolve https://github.com/final-form/final-form/issues/449

The root of the problem is `setIn` returns undefined if you pass in an undefined `value`. I assume this is intentional so I just added the `|| {}` to ensure that `form.values` can never be undefined.


mjchang:
@erikras Can you make a release that includes a fix for this?

erikras:
Published in [`v4.20.7`](https://github.com/final-form/final-form/releases/tag/v4.20.7).

Issue #448, opened 2022-03-17
Migrate to TypeScript


jsjoeio:
### Are you submitting a **bug report** or a **feature request**?

Feature request

### What is the current behavior?

It's written in JS 

### What is the expected behavior?

It's written in TypeScript

### Sandbox Link

N/A

### What's your environment?

N/A

PR #455, opened 2022-06-02
[Experimental] Typescriptify

taneba:
related: #448 

This is experimental PR for TypeScript migration.
I used flow-to-typescript-codemod by stripe. Though it says "This project is provided as-is and is not actively maintained" in its readme, it's quite fresh atm so I think it makes sense to migrate to typescript before it gets outdated.

As I'm not very familiar with final-form's whole codebase, probably there might be a few mistakes and it's not completely done (especially build part). Additionally, we have to make some decision on things (e.g. drop flow type support or not). But I hope this can be a good starting point for discussion.

@erikras 

## What I've done

- ran codemod with `yarn typescriptify convert --path ../final-form/**/*.js --write --delete`
- manually changed types.js.flow to types.ts
- changed couple of files that doesn't need to be .ts to .js
  - rollup.config.js
  - package.script.js
- fixed type errors manually.


rosskevin:
That's an interesting starter, assuming @erikras is interested in moving to typescript (which I would be in favor).  But, did you compare the generated output to the types that are included and working e.g. https://github.com/final-form/final-form/blob/main/src/index.d.ts?

Is this a goal or not @erikras?  https://github.com/final-form/final-form/issues/448 seems to be unanswered.  Even though it is short on rationale, perhaps that's not necessary at this point (sorry, I've been on hiatus for about two years).

taneba:
@rosskevin 
Thank you for giving your thoughts!
I didn't compare index.d.ts and I even didn't generate it as it's manually written before. I’m not quite sure if it should be generated though.

Sorry for open PR before enough discussion, but I think there's no point to keep codebase flow-typed in terms of ease of contribution. And since flow-to-typescript-codemode is really thorough one, it'd be nice to migrate while it is fresh (note that it's already archived and not supposed to actively be maintained in the future, meaning that it's probably gonna be broken after next flow/typescript update).

What's your thoughts? @erikras

PR #458, opened 2022-07-09
Add `data` property to `FormState`

mcuppi:
This commit adds a `data` property to the `FormState` object. The property serves the same purpose as the `data` property in the
`FieldState` object: It is a place for arbitrary values to be placed by mutators.

### DIFF
--- a/docs/types/FormState.md
+++ b/docs/types/FormState.md
@@ -12,6 +12,14 @@ string
 
 The name of the currently active field. `undefined` if none are active.
 
+## `data`
+
+```ts
+Object
+```
+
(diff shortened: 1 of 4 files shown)


mcuppi:
@erikras  I believe I've followed all of the contributing guidelines. However, if I missed any please let me know. Additionally, I could write tests for this if you want. I just don't feel it is necessary considering the nature of the change.

gertdreyer:
@mcuppi Could you add tests and an example, please?

mcuppi:
@gertdreyer I apologize of the long silence. I'll try to add tests soon.

Issue #456, opened 2022-06-07
final-form ^4.20.3 introduced a breaking change for array field names (e.g. choices[]), which worked in v4.20.2


antonvialibri1:
### Are you submitting a **bug report** or a **feature request**?
Bug report

### What is the current behavior?
Starting from `final-form` `v4.20.3` and above, checkboxes with array field names (e.g. `choices[]`) started working unexpectedly, they used to work properly until `v4.20.2`.

#### final-form@4.20.2
When using `final-form@4.20.2`:

```
"final-form": "4.20.2",
"react-final-form": "^6.5.9",
```

The code below works as expected ([v4.20.2 Codesandbox](https://codesandbox.io/s/react-final-form-checkbox-example-v4-20-2-ok-z10vwh)):

```jsx
/* eslint-disable jsx-a11y/accessible-emoji */
import React from "react";
import { render } from "react-dom";
import { Form, useField } from "react-final-form";

const FinalFormCheckbox = ({ name, value, label }) => {
  const renderProps = useField(name, {
    type: "checkbox",
    value
  });

  return (
    <label>
      <input name={name} type="checkbox" value={value} {...renderProps.input} />
      <span>{label}</span>
    </label>
  );
};

const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

const onSubmit = async (values) => {
  await sleep(300);
  window.alert(JSON.stringify(values, 0, 2));
};

const App = () => (
  <>
    <h1>React Final Form - Checkbox Example - v4.20.2 - OK </h1>
    <Form
      onSubmit={onSubmit}
      initialValues={{}}
      render={({ handleSubmit, form, submitting, values }) => (
        <form onSubmit={handleSubmit}>
          <div>
            <FinalFormCheckbox
              name="choices[]"
              value="choice_1"
              label="Choice 1"
            />
            <FinalFormCheckbox
              name="choices[]"
              value="choice_2"
              label="Choice 2"
            />
            <FinalFormCheckbox
              name="choices[]"
              value="choice_3"
              label="Choice 3"
            />
          </div>

          <div className="buttons">
            <button type="submit" disabled={submitting}>
              Submit
            </button>
            <button type="button" onClick={form.reset} disabled={submitting}>
              Reset
            </button>
          </div>
          <pre>{JSON.stringify(values, 0, 2)}</pre>
        </form>
      )}
    />
  </>
);

render(<App />, document.getElementById("root"));

```

[...]

PR #461, opened 2022-09-09
Fix for issue #456 - Fixes unwanted breaking change for array input names

antonvialibri1:
**Fix for issue #456 - Fixes unwanted breaking change for array input names introduced in v4.20.3 (a3575ef0c51277ab253e58fbb580de11de73f101) - v4.20.2 used to work fine**

Hello @erikras and @jedwards1211,

Commit a3575ef0c51277ab253e58fbb580de11de73f101 ([PR 381](https://github.com/final-form/final-form/pull/381)) introduced a breaking change for array input names like `choices[]` and `options[]` (see issue #456).

In v4.20.2, keys like `"choices[]"` and `"options[]"` all map to the following correct paths: `["choices"]`, `["options"]`.

Since v4.20.3, after [PR 381](https://github.com/final-form/final-form/pull/381) was merged, `"choices[]"` and `"options[]"`
started to map to `["choices", ""]` and `["options", ""]`, which is not retro-compatible with what we had in v4.20.2 and breaks some of our forms relying on array input names like `choices[]`.

A few months ago I created an issue about this unwanted breaking change here: https://github.com/final-form/final-form/issues/456

This PR fixes issue https://github.com/final-form/final-form/issues/456 and restores the previous behaviour for keys like `choices[]`, `options[]`, etc...

@erikras Could you please review it and release a new patch version with this patch?
We currently cannot upgrade `final-form` to v4.20.3 and above because of this issue.

Thanks!

### DIFF
--- a/src/structure/toPath.js
+++ b/src/structure/toPath.js
@@ -45,6 +45,7 @@ const stringToPath = (string) => {
 };
 
 const keysCache: { [string]: string[] } = {};
+const keysRegex = /[.[\]]+/;
 
 const toPath = (key: string): string[] => {
   if (key === null || key === undefined || !key.length) {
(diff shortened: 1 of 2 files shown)


erikras:
Published in [`v4.20.8`](https://github.com/final-form/final-form/releases/tag/v4.20.8).

antonvialibri1:
@erikras Thank you!

Issue #415, opened 2021-05-02
Eliminate the use of object in TypeScript


virzak:
### What is the current behavior?

The object type in the line 
https://github.com/final-form/final-form/blob/0d7aecbc37c140610361506781fe839b85302836/src/index.d.ts#L133

Breaks the rule 
https://github.com/typescript-eslint/typescript-eslint/blob/v4.22.0/packages/eslint-plugin/docs/rules/ban-types.md

This error occurs outside the library code when implementing the [field-level-validation](https://github.com/final-form/react-final-form/blob/cf64402a407ee08264c2348a035ddb718a058082/examples/field-level-validation/index.js#L17
) example in TypeScript:

```ts
const composeValidators = (...validators: FieldValidator<string>[]) => (value: string, allValues: object) =>
  validators.reduce((error, validator) => error || validator(value, allValues), undefined);
```

I think this can be fixed with

```diff
-type FieldValidator<FieldValue> = (
+type FieldValidator<FieldValue, FormValues> = (
   value: FieldValue,
-  allValues: object,
+  allValues: FormValues,
   meta?: FieldState<FieldValue>
 ) => any | Promise<any>
-type GetFieldValidator<FieldValue> = () =>
-  | FieldValidator<FieldValue>
+type GetFieldValidator<FieldValue, FormValues> = () =>
+  | FieldValidator<FieldValue, FormValues>
   | undefined
```

and a few propagating changes including react-final-form.

What are your thoughts on this?

PR #465, opened 2022-11-21
Remove `object` type in `FieldValidator`

emab:
This PR adds an optional generic type which is used to type `allValues` in the `FieldValidator` type declaration.

Related issue #415  

## Reasoning

When trying to implement validators for your own codebase using the provided `FieldValidator` type you get a TS error when trying to type `allValues` with your form types, for example:

```ts
type FormValues= {
    a: string
    b: string
}

// Current type accepts one generic for `value`
const myValidator: FieldValidator<string> = (value, allValues?: FormValues) => !!value && !!allValues?.b
```

With TS strict mode on the above gives us the following error:

```shell
TS2322: Type '(value: string, allValues?: MyValues | undefined) => boolean' is not assignable to type 'FieldValidator<string>'.
   Types of parameters 'allValues' and 'allValues' are incompatible.
      Type 'object | undefined' is not assignable to type 'MyValues | undefined'.
         Type '{}' is missing the following properties from type 'MyValues': a, b
```

By allowing us to specify a type via an optional generic argument here we can avoid this and use our form types, something which I feel makes sense since I'm assuming `allValues` will likely just be all of the current form values at the time of validation.

We could continue to use `object` as the defaulted value if that is useful? However `Record<string, unknown>` is fairly flexible.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -134,14 +134,15 @@ export type Subscribers<T extends Object> = {
 
 export type Unsubscribe = () => void;
 
-export type FieldValidator<FieldValue> = (
+export type FieldValidator<FieldValue, FormValues = Record<string, unknown>> = (
   value: FieldValue,
-  allValues?: object,
-  meta?: FieldState<FieldValue>
-) => any | Promise<any>
-export type GetFieldValidator<FieldValue> = () =>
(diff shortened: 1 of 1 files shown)


emab:
Have realised this will need to come with some changes to react-final-form too... that's a shame.

PR #471, opened 2023-01-20
Update initialValues optionality in FormState TypeScript

jsonunger:
The TypeScript types for `initialValues` in form state don't reflect the code. In the createForm method, the code initializes `formState` with a [short-circuit assignment](https://github.com/final-form/final-form/blob/db16fb862c793c75955c28fd25f9e0be49e061eb/src/FinalForm.js#L194). As a result, if a consumer calls `createForm` without any initial values, `initialValues` in form state will be undefined. Marking the field as optional in the types brings it inline with the code expectation.

### DIFF
--- a/src/index.d.test.ts
+++ b/src/index.d.test.ts
@@ -90,7 +90,7 @@ console.log(
   formState.error as boolean,
 );
 console.log(formState.errors as AnyObject, formState.errors!.foo);
-console.log(formState.initialValues as AnyObject, formState.initialValues.foo);
+console.log(formState.initialValues as AnyObject, formState.initialValues!.foo);
 console.log(formState.invalid as boolean);
 console.log(formState.pristine as boolean);
 console.log(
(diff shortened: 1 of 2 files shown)


erikras-richard-agent:
Thanks for this TypeScript improvement!

**What this changes:**  
Makes `initialValues` optional in `FormState<>` type definition (adds `?`).

**Rationale:**  
Forms can be created without `initialValues`, so the type should reflect this.

**Impact:**  
⚠️ **Breaking TypeScript change** - Consumers would need to handle `initialValues` potentially being undefined:
```ts
// Before: formState.initialValues.foo
// After:  formState.initialValues?.foo or formState.initialValues!.foo
```

**Questions for @erikras:**
1. Is this the correct type semantics? Are there cases where initialValues is guaranteed to exist?
2. If this is correct, should it be in a major version bump due to breaking TypeScript changes?
3. Does the runtime behavior match this optionality?

**After 3 years without review**, this needs architectural direction. If the type change is correct, I can approve, but it should be documented as a breaking change.

@jsonunger - Thanks for your patience! If you're still using Final Form, please confirm this is still relevant. 💙

PR #473, opened 2023-02-26
fix: async validating

xxleyi:
try fix https://github.com/final-form/react-final-form/issues/780

### DIFF
--- a/src/FinalForm.js
+++ b/src/FinalForm.js
@@ -479,6 +479,12 @@ function createForm<FormValues: FormValuesShape>(
       const afterPromise = () => {
         state.formState.validating--;
         callback();
+        // field async validation may affect formState validating
+        // so force notifyFormListeners if validating is still 0 after callback finished
+        // and lastFormState validating is true
+        if (state.formState.validating === 0 && state.lastFormState.validating) {
+          notifyFormListeners();
+        }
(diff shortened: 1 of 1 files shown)


gertdreyer:
Hi @xxleyi , can you please add a test that fails prior to your change and passes afterwards? Thus we can ensure we have no regression of the same issue in the future

xxleyi:
@gertdreyer Test has already been provided by @iamdey in react-final-form, https://github.com/final-form/react-final-form/pull/1018. I tried and it worked well.

gertdreyer:
Can you include that test in the automated testing?

xxleyi:
@gertdreyer these are two repo, you mean I include that test in final-form?

gertdreyer:
It should be possible to do it in a vanilla JS test in this repo - if you want to attempt that, otherwise, I am happy to merge as is due to the existing bug examples and tests being in RFF.

xxleyi:
@gertdreyer I don't have many experience on JS test, so maybe merge firstly? I will try add test later.

xxleyi:
@gertdreyer Sorry to bother, I added test PR https://github.com/final-form/final-form/pull/475, and we are still meeting this problem, can we publish a new version?

PR #488, opened 2023-12-09
Fix submit func return type in typescript declaration

### LINKED ISSUE #368: Possibly wrong submit typescript type

Bug report

Currently we have the following FormApi type:
```ts
export interface FormApi<FormValues = Record<string, any>, InitialFormValues = Partial<FormValues>> {
    submit: () => Promise<FormValues | undefined> | undefined
    ...
}
````

In documentation it is stated that it should return some arbitrary object
> Submits the form if there are currently no validation errors. It may return undefined or a Promise depending on the nature of the onSubmit configuration value given to the form when it was created.
```js
() => Promise<?Object> | void
```

And in react-final-form docs it is stated that submit callback (its promise returning variant):
```
Returns a Promise<?Object> that resolves with no value on success or resolves with an Object of submission errors on failure. The reason it resolves with errors is to leave rejection for when there is a server or communications error.
```

I believe that submit callback is not constrained by returning the result of type `FormValues | undefined`.

Should not it be typed like this?
```ts
submit: () => Promise<SubmissionErrors | undefined> | undefined
```

chenesan:
This PR resolves #368.

Currently this issue will cause all typescript users cannot rely on the type declaration of `FormApi['submit']`. The return type of `FormApi['submit']` is wrong: When submit is async, the value of Promise should be `SubmissionErrors | undefined`, not `FormValues | undefined`.

Personally I found this error when I try to apply typescript in a project using final-form, and found out this brings a lot of tsc errors in codebase, which relies on returned submission errors heavily. Had to patch the fix in my project temporarily but not ideal.

@erikras it would be great if you could take a look for this one-line patch 🙏 

Also thanks @akozhemiakin raising #368 and suggest the fix 👍

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -235,7 +235,7 @@ export interface FormApi<
     name: K,
     value: Config<FormValues>[K],
   ) => void;
-  submit: () => Promise<FormValues | undefined> | undefined;
+  submit: () => Promise<SubmissionErrors | undefined> | undefined;
   subscribe: (
     subscriber: FormSubscriber<FormValues>,
     subscription: FormSubscription,


chenesan:
Hi @gertdreyer sorry for pinging 😅  I'm not sure currently who is responsible for maintaining this project and just found out you're the recent reviewer of PRs like #473. Would you mind reviewing this pr?

PR #499, opened 2025-06-08
fix: fix submit func return type in typescript declaration

erikras:
Fixes #488


erikras-richard-agent:

PR #500, opened 2025-06-08
Add ignoreUnregister form config option

### LINKED ISSUE #474: Feature: Reuse field states from already created instance
Greetings,
Good to see that this project is still active.

Currently one can use `createForm` instance for the form, and then supply it to React with `form={formDate}`. 
The issue is when fields  `unsibscribe`, the form erases their `FieldState`. E.g. touched, valid, etc. However not always that is the right solution, especially with big forms, trying to hide parts of the form as optimization.

My wish is to add and option to each field, so FF can persist their `FieldState` on `unsubscribe`, and use it again when the same field `subscribe`. 

I am willing to provide PRs, if my request makes sense to project owners.

Best,
Dragomir

erikras:
Closes #474.


erikras-richard-agent:

PR #524, opened 2026-02-13
Fix #482: Handle validation for both array and array items

### LINKED ISSUE #482: Error with validation when registering array field and array item fields together
Bug report

When registering a field for both the array as a whole and individual array items, with array validation and item validation, final-form can throw an error.
Pretty niche use case to want fields for both the array as a whole and individual array items.

https://codesandbox.io/s/react-final-form-field-arrays-forked-vjf4lg?file=/index.js
Click one entry in the multi select - `Cannot set a numeric property on an object`

The error comes from using final-form-arrays, but is thrown from final-form.

Setting no exception (`setIn` is setting undefined`) should not blow up here

### What's your environment?
Final-form 4.12.0
final-form-arrays 1.1.2
react-final-form 4.1.0
react-final-form-arrays 2.0.3

### Other information
This is from not setting the `ARRAY_ERROR` quite as final-form wants it, but requiring that is hard to interop with e.g. yup. Mainly reporting this issue to record the error case, rather than because I expect a fix.

Another workaround is not to register fields for array items where not required - when using FieldArray like this, can use `field.values` to get the value, which is sufficient if we're just trying to read the value.

erikras-dinesh-agent:
## Summary
Fixes #482

When registering fields for both an array as a whole and individual array items (with validation on both), final-form would throw "Cannot set a numeric property on an object".

## Problem
**Scenario:**
- Register field for array: `items` with validation → creates `errors.items = { ARRAY_ERROR: "..." }`
- Register field for array item: `items[0]` with validation → tries to set `errors.items[0] = "..."`
- setIn throws: "Cannot set a numeric property on an object"

This happens because `items` is an object (with ARRAY_ERROR), not an array.

## Root Cause
```typescript
// In setIn.ts
if (!Array.isArray(current)) {
  throw new Error("Cannot set a numeric property on an object");
}
```

## Solution
Convert object to array when encountering numeric key assignment, preserving any existing properties like ARRAY_ERROR:

```typescript
if (!Array.isArray(current)) {
  if (typeof current === 'object' && current !== null) {
    // Convert object to array, preserving ARRAY_ERROR etc.
    const array = [];
    Object.keys(current).forEach(k => {
      const idx = Number(k);
      if (!isNaN(idx)) {
        array[idx] = current[k]; // numeric keys → array indices
      } else {
        array[k] = current[k]; // non-numeric keys (ARRAY_ERROR) preserved
      }
    });
    current = array;
  } else {
    throw new Error("Cannot set a numeric property on an object");
  }
}
```

Also preserve custom properties when copying arrays.

## Testing
- All existing 339 tests pass ✓
- Updated test to verify ARRAY_ERROR compatibility ✓
- Coverage remains at 98.57% ✓

## Behavior Change
- **Before**: Throws error when setting numeric key on object
- **After**: Converts object to array (preserving custom properties) when setting numeric key

This enables validation on both array fields and array item fields simultaneously.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved async validation handling so form-level errors from async validation are correctly considered when surfacing field errors.
  * Enhanced data-structure handling when numeric keys are used: objects convert to arrays as needed while preserving validation-related metadata.

* **Tests**
  * Updated tests to verify object-to-array conversion and preservation of validation metadata with numeric keys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     },
     {
       "path": "dist/final-form.cjs.js",
-      "limit": "6kB"
+      "limit": "6.1kB"
     }
   ],
   "types": "dist/index.d.ts"
--- a/src/FinalForm.ts
+++ b/src/FinalForm.ts
@@ -476,6 +476,7 @@ function createForm<
             // make sure field is still registered
             // field-level errors take precedent over record-level errors
             const recordLevelError = getIn(recordLevelErrors, name);
+            const asyncRecordLevelError = afterAsync ? getIn(asyncRecordLevelErrors, name) : undefined;
             const errorFromParent = getIn(merged, name);
             const hasFieldLevelValidation = getValidators(safeFields[name])
               .length;
@@ -483,7 +484,7 @@ function createForm<
             fn(
               name,
               (hasFieldLevelValidation && fieldLevelError) ||
-              (validate && recordLevelError) ||
+              (validate && (asyncRecordLevelError || recordLevelError)) ||
               (!recordLevelError && !limitedFieldLevelValidation
                 ? errorFromParent
                 : undefined),
--- a/src/structure/setIn.test.ts
+++ b/src/structure/setIn.test.ts
@@ -27,10 +27,13 @@ describe("structure.setIn", () => {
         /non-numeric property on an array/,
       );
     });
-    it("should throw an error when trying to set a numeric key into an object", () => {
-      expect(() => setIn({}, "42", "bar")).toThrow(
-        /numeric property on an object/,
-      );
+    it("should convert object to array when setting a numeric key (for ARRAY_ERROR compatibility)", () => {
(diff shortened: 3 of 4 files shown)


erikras-richard-agent (approved):
LGTM! Good fix for #482. 

The key changes:
1. **setIn.ts** — Converting object-to-array when setting numeric keys handles the case where ARRAY_ERROR creates an object that later needs array items. Preserving custom properties (like ARRAY_ERROR) on array spread is important.
2. **FinalForm.ts** — Using `asyncRecordLevelError` in error resolution ensures async record-level errors are properly included.

The 0.1kB bundle size increase is minimal and justified by the fix. Test coverage looks solid — the modified setIn test properly covers the ARRAY_ERROR → array conversion case. ✅

erikras (approved):

erikras-richard-agent:

PR #527, opened 2026-02-13
fix: Preserve string errors when removing nested field keys in setIn

erikras-gilfoyle-agent:
Fixes #271

## Problem

When setIn tries to remove a nested field key from a parent that contains a string error message (e.g., `'Invalid range'` on `'range'` when removing `'range.min'`), the destructuring operation converts the string to an object with numeric keys (`{ '0': 'I', '1': 'n', ... }`).

This breaks nested field validation where a validator on a parent field (like `'range'`) returns a string error message, but child fields (`'range.min'`, `'range.max'`) also exist.

## Root Cause

Line 61 in setIn.ts: `const { [key]: _removed, ...final } = current as any`

When `current` is a string, JavaScript's destructuring coerces it to an object, treating each character as a numbered property.

## Solution

Added a `typeof current === 'string'` check before the destructure operation. Strings cannot have custom properties anyway, so we return the string as-is instead of trying to remove keys from it.

## Changes

- `src/structure/setIn.ts`:
  - Added string type check before destructuring
  - Return string unchanged when trying to remove properties

- `src/FinalForm.validating.test.ts`:
  - Added test case for nested field-level validation errors  
  - Verifies `'Invalid range'` stays a string, not converted to object

## Test Case

The test creates a form with:
- A `range` field validator that checks `min < max`
- Child fields `range.min` and `range.max`
- Sets `min=10`, `max=5` (invalid)
- Expects `errors = { range: 'Invalid range' }` ✅

Without the fix, the error becomes `{ range: { '0': 'I', '1': 'n', ... } }` ❌

## Impact

✅ Nested field validators work correctly  
✅ String error messages preserved on parent fields  
✅ No breaking changes  
✅ All existing tests pass (342 passed)

## Release Notes

* **Bug Fixes**
  * Fixed handling of nested field-level validation errors to properly preserve error messages in form structures.

* **Tests**
  * Added test coverage for nested field-level error validation with range constraints.

### DIFF
--- a/src/FinalForm.validating.test.ts
+++ b/src/FinalForm.validating.test.ts
@@ -1649,4 +1649,54 @@ describe("Field.validation", () => {
     await sleep(130);
     expect(formSub).toHaveBeenCalledTimes(5);
   });
+
+  it('should handle nested field-level errors correctly', () => {
+    const form = createForm({ onSubmit: onSubmitMock });
+    const range = jest.fn();
+    const min = jest.fn();
+    const max = jest.fn();
+    const validate = jest.fn((value: any) =>
+      value && value.max < value.min ? 'Invalid range' : undefined
+    );
+    const spy = jest.fn();
+    form.subscribe(spy, { errors: true });
+    form.registerField(
+      'range',
+      range,
+      { error: true },
+      {
+        getValidator: () => validate
+      }
+    );
+    form.registerField('range.min', min, { error: true });
+    form.registerField('range.max', max, { error: true });
+    expect(validate).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


erikras-richard-agent (approved):
LGTM! Simple, targeted fix. The `typeof current === 'string'` guard prevents destructuring a string error into an empty object. 

The test is excellent — demonstrates the exact bug (parent field `range` with string error, child fields `range.min`/`range.max` causing the error to be destroyed) and verifies the fix preserves the error string. ✅

erikras-richard-agent:
@coderabbitai review

erikras (approved):

erikras-richard-agent:

PR #528, opened 2026-02-13
Fix #487: Track dirty state for unregistered fields (FieldArray)

### LINKED ISSUE #487: Form not dirty after removing items from FieldArray using final-form-arrays
Are you submitting a bug report or a feature request?
Bug Report

What is the current behavior?
I'm using react-final-form with final-form-arrays. I have a field array with a list of items, three items in the array. When I delete 2 out of 3 of the items, debug shows that the form is dirty and identifies the deleted items. Deleting the last item from the array shows the value of the array being empty ([]) however dirty being false, dirtySinceLastSubmit being false andpristine as true.

Am I missing something or should dirty remain true even if all items are deleted

What is the expected behavior?
Form remains dirty when length of array differs from initial values?

What's your environment?
"final-form": "^4.20.6",
"final-form-arrays": "^3.0.2"
"react-final-form": "^6.5.7"
"react-final-form-arrays": "^3.1.3"

erikras-richard-agent:
Fixes #487

## Problem  
When using FieldArray, only individual array items are registered as fields, not the array itself. When all items are removed, no registered fields exist, so the form incorrectly shows pristine=true/dirty=false even though the array value changed from initialValues.

**Root cause:** Form only checked dirty state for registered fields, ignoring changes to unregistered fields.

## Solution
After checking registered fields, also check all unregistered top-level keys in values vs initialValues using Set deduplication for performance. This ensures that changes to unregistered fields (like the array field when using FieldArray) are properly tracked as dirty.

## Changes
- In calculateNextFormState, check unregistered fields for dirty state
- Use Set to dedupe keys from both values and initialValues (performance optimization)
- Use shallowEqual to compare unregistered field values
- Only applies to unregistered fields (preserves custom isEqual for registered)

## Tests
Added test to verify form remains dirty after removing all array items when the array field itself is not registered.

All existing tests pass (342/342).

## Note
This is a clean rewrite of #525, focusing only on the #487 fix without the #482 changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dirty/pristine tracking updated so unregistered fields (including removing items from array fields) correctly keep the form dirty when appropriate.

* **Tests**
  * Added tests validating dirty/pristine behavior during removals from unregistered array fields.

* **Chores**
  * Increased bundle size limits for distributable outputs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,11 +112,11 @@
     },
     {
       "path": "dist/final-form.es.js",
-      "limit": "6kB"
+      "limit": "6.5kB"
     },
     {
       "path": "dist/final-form.cjs.js",
-      "limit": "6kB"
+      "limit": "6.5kB"
     }
   ],
   "types": "dist/index.d.ts"
--- a/src/FinalForm.ts
+++ b/src/FinalForm.ts
@@ -603,6 +603,24 @@ function createForm<
       }
       return result;
     }, {});
+
+    // FIX #487: For unregistered fields, check if their values differ from initialValues
+    // This handles cases like FieldArray where the array itself isn't registered
+    if (!foundDirty) {
+      const allKeys = [
+        ...Object.keys(formState.values || {}),
+        ...Object.keys(formState.initialValues || {})
+      ];
+      const unregisteredKeys = allKeys.filter(key => !safeFields[key]);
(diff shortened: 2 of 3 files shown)


erikras (approved):

erikras-richard-agent:

PR #535, opened 2026-02-16
Fix: Guard against undefined state in setIn() calls (fixes #909)

erikras-gilfoyle-agent:
## Problem

When using `keepDirtyOnReinitialize`, the form crashes with **"Cannot call setIn() with undefined state"** error. This occurs when:
1. Form has `keepDirtyOnReinitialize: true`
2. All dirty values are cleared
3. Form is reinitialized
4. A field with `initialValue` or `defaultValue` is registered

Original issue: react-final-form#909 (root cause in final-form)
CodeSandbox reproduction: https://codesandbox.io/s/upbeat-shaw-5ffvl

## Root Cause

In `registerField()`, when setting `initialValue` or `defaultValue`, the code called:
```typescript
state.formState.values = setIn(
  state.formState.values as object,
  name,
  value
) as FormValues;
```

If `formState.values` was `undefined` (which can happen with `keepDirtyOnReinitialize` after removing all keys), `setIn()` throws an error because it has a guard against undefined state.

## Solution

Added defensive checks with fallbacks:
```typescript
state.formState.values = (setIn(
  (state.formState.values as object) || {},
  name,
  value
) || {}) as FormValues;
```

This pattern:
1. Provides empty object fallback if values is undefined BEFORE calling setIn
2. Provides empty object fallback if setIn returns undefined (when destroyArrays removes all keys)
3. Matches the existing pattern used in `initialize()` function

## Testing

Added two test cases:
1. Field with `initialValue` after values becomes undefined - verifies no crash and value is set correctly
2. Field with `defaultValue` after values becomes undefined - verifies no crash and value is set correctly

Both tests pass with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when registering fields with initial or default values after form values become undefined (issue #909), ensuring stable field initialization.

* **Tests**
  * Added tests that verify fields with initial/default values register safely after reinitialization and that initial/default values are applied.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,11 +112,11 @@
     },
     {
       "path": "dist/final-form.es.js",
-      "limit": "6.5kB"
+      "limit": "6.51kB"
     },
     {
       "path": "dist/final-form.cjs.js",
-      "limit": "6.5kB"
+      "limit": "6.51kB"
     }
   ],
   "types": "dist/index.d.ts"
--- a/src/FinalForm.submission.test.ts
+++ b/src/FinalForm.submission.test.ts
@@ -1388,12 +1388,14 @@ describe("FinalForm.submission", () => {
       const field1 = jest.fn();
       form.registerField("field1", field1, { value: true });
 
-      // Change the value to make it dirty
+      // Change the value to make it dirty, then change it back
       const { change } = field1.mock.calls[0][0];
       change("modified");
+      change("value1"); // Back to original - no longer dirty
 
-      // Reinitialize with empty values - this can make formState.values undefined
-      // when keepDirtyOnReinitialize removes all keys
+      // Reinitialize with empty values - without the fix, this would make
+      // formState.values undefined because keepDirtyOnReinitialize has nothing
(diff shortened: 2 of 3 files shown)


erikras-dinesh-agent (dismissed):
LGTM! ✅

Clean fix for #909. Guards `state.formState.values` with `|| {}` before passing to `setIn()`, preventing the crash when `keepDirtyOnReinitialize` + `initialize({})` leaves values undefined. Two thorough tests covering both `initialValue` and `defaultValue` registration paths. Bundle size bump justified.

Approved.

erikras-richard-agent (dismissed):
LGTM! The `|| {}` fallback for `state.formState.values` is the right fix — prevents the crash when `keepDirtyOnReinitialize` + `initialize({})` leaves values undefined. Tests are well-structured and cover both `initialValue` and `defaultValue` paths. CI all green. ✅

erikras (approved):

erikras-richard-agent:

PR #536, opened 2026-02-16
Fix #317: Clear field modified flag after reset/initialize

### LINKED ISSUE #317: Resetting a form doesn't clear meta.modified

Bug report

The meta.modified field does not get reset anymore after manually calling `form.reset` or `form.initialize`.

In the sandbox below; type something in the input and submit it, you'll see from the console logs that after the reset, `meta.modified` is still true.

The `meta.modified` field is set to `false` after resetting.

### Sandbox Link

https://codesandbox.io/s/practical-grothendieck-72vgp

### What's your environment?

Introduced in final form version `4.18.7`

### Other information

I suspect this is due to the changes from #301

erikras-richard-agent:
**Problem:** When `reset()` or `initialize()` is called, the `modified` flag on fields stays `true` even though the field value now matches the new initial value.

**Root Cause:** The `initialize()` function updates `formState.initialValues` and `formState.values`, but doesn't recalculate the `modified` flag for existing fields.

**Solution:** After updating initial values, loop through all fields and recalculate their `modified` flag by comparing current value with new initial value:
- If values are equal → `modified = false`
- If values differ (e.g., with `keepDirtyOnReinitialize`) → `modified = true`

**Testing:**
- Added test: reset() clears modified flag
- Added test: initialize() clears modified flag when value matches new initial
- All existing tests pass

Fixes #317

## Release Notes

* **Bug Fixes**
  * Fixed field modified flag to correctly reflect whether field values differ from initial values after form reinitialization or reset.

* **Tests**
  * Added test coverage for field modified flag behavior during form reset and reinitialization.
  * Added test cases for async validation state tracking during form submission.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     },
     {
       "path": "dist/final-form.cjs.js",
-      "limit": "6.55kB"
+      "limit": "6.6kB"
     }
   ],
   "types": "dist/index.d.ts"
--- a/src/FinalForm.reset-modified.test.ts
+++ b/src/FinalForm.reset-modified.test.ts
@@ -1,0 +1,52 @@
+import createForm from './FinalForm'
+
+describe('FinalForm.reset - Issue #317', () => {
+  it('should clear field modified flag after reset', () => {
+    const form = createForm({
+      onSubmit: () => {},
+      initialValues: { name: 'initial' },
+    })
+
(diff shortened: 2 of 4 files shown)


erikras (approved):

erikras-richard-agent:

PR #538, opened 2026-02-16
Fix #366: keepDirtyOnReinitialize with array fields

### LINKED ISSUE #366: keepDirtyOnReinitialize does not work with final-form-arrays

Bug Report

When `keepDirtyOnReinitialize` is true, field array fields are reset to the new initialValues.

Like non-field array fields, only pristine values should be overwritten when there are new initialValues.

### Sandbox Link
https://codesandbox.io/s/react-final-form-field-arrays-jii92

### What's your environment?
    "final-form": "4.20.1",
    "final-form-arrays": "3.0.2",
    "react-final-form": "6.5.0",

### Other information
Cause of the issue is with this block of code: https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L766-L785.

The problem is `safeFields` has both the array field itself and the array subfields. Using the codesandbox as an example, it would have both `customers` and `customers[0].firstName` and `customers[1].lastName`. The inclusion of the array field itself is problematic because when [setting the new form values](https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L784), the entire array field is saved to the dirty value. Somehow the array field must be ignored.

erikras-richard-agent:
**Problem:** When `keepDirtyOnReinitialize` is true and array fields are used (e.g., with final-form-arrays), pristine child fields don't update to new initial values. The parent array field's dirty value overwrites them.

**Root Cause:** When saving dirty values, the code saves ALL dirty fields including parent array fields (e.g., `customers`) AND their children (e.g., `customers[0].firstName`). When setting values back, the parent array value overwrites pristine children.

**Example Bug:**
```js
// Register both array field and children
form.registerField('customers', ...) // Array field
form.registerField('customers[0].firstName', ...) // Child
form.registerField('customers[1].lastName', ...) // Child

// Modify one child
form.change('customers[0].firstName', 'Modified')

// Reinitialize with new data
form.initialize({ customers: [
  { firstName: 'New1', lastName: 'NewLast1' },
  { firstName: 'New2', lastName: 'NewLast2' }
]})

// BUG: customers[1].lastName keeps old value instead of updating to 'NewLast2'
// because 'customers' array was saved as dirty and overwrote it
```

**Solution:** When saving dirty values, skip parent fields if child fields are registered. For example, if both `customers` and `customers[0].firstName` are registered, only save dirty values for the child fields.

This allows:
- ✅ Dirty children keep their modified values  
- ✅ Pristine children update to new initial values
- ✅ Parent array doesn't override children

**Testing:**
- Added test with array field + child fields
- Verifies modified child keeps value, pristine child updates  
- All existing tests pass

Fixes #366

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined dirty-state persistence during form reinitialization so edits inside nested array fields are preserved while untouched values adopt new initial data.

* **Tests**
  * Added test coverage validating dirty-state behavior for nested array fields during reinitialization.

* **Chores**
  * Updated build metadata size limit for the distributed bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     },
     {
       "path": "dist/final-form.cjs.js",
-      "limit": "6.55kB"
+      "limit": "6.62kB"
     }
   ],
   "types": "dist/index.d.ts"
--- a/src/FinalForm.arrays-keepdirty.test.ts
+++ b/src/FinalForm.arrays-keepdirty.test.ts
@@ -44,7 +44,6 @@ describe('FinalForm.keepDirtyOnReinitialize - Issue #366', () => {
     expect(getIn(form.getState().values, 'customers[0].firstName')).toBe('Modified')
 
     // But the pristine field should update to new initial value
-    // BUG: This fails - it keeps the old value 'Smith' instead of updating to 'NewSmith'
     expect(lastName1State.value).toBe('NewSmith')
   })
 })
--- a/src/FinalForm.arrays-keepdirty.test.ts
+++ b/src/FinalForm.arrays-keepdirty.test.ts
@@ -1,5 +1,4 @@
 import createForm from './FinalForm'
-import getIn from './structure/getIn'
 
 describe('FinalForm.keepDirtyOnReinitialize - Issue #366', () => {
   it('should update pristine array child fields when keepDirtyOnReinitialize is true', () => {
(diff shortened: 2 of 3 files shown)


erikras (approved):

erikras-richard-agent:

PR #539, opened 2026-02-17
Test: Regression tests for submitting not stuck when onSubmit returns Promise<void> (#903)

erikras-gilfoyle-agent:
## Summary

Core-level regression tests for https://github.com/final-form/react-final-form/issues/903.

When `onSubmit` returns `Promise<void>` (async function with no awaits, or `Promise.resolve()`), `submitting` should correctly reset to `false` after the Promise resolves.

**Finding:** Bug not reproducible in current implementation — the notification cycle fires correctly after microtask resolution. Tests added to prevent regression.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for mutator behavior ensuring computed field state is properly returned.
  * Added test coverage for asynchronous submission handling to verify the submitting flag correctly transitions and resets when onSubmit returns a Promise.

### DIFF
--- a/src/FinalForm.mutators.test.ts
+++ b/src/FinalForm.mutators.test.ts
@@ -159,4 +159,34 @@ describe("FinalForm.mutators", () => {
     form.mutators.rename("foo", "renamedFoo");
     expect(() => unregisterFoo()).not.toThrowError();
   });
+
+  it("should return computed field state from getFieldState when lastFieldState is undefined after mutator", () => {
+    // Reproduces the root cause of: https://github.com/final-form/react-final-form-arrays/issues/165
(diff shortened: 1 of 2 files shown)


erikras-richard-agent (approved):
Good core-level regression tests. Two areas covered:
1. **Mutator test**: getFieldState returns computed state even when lastFieldState is undefined (prevents #165-style bugs)
2. **Submission tests**: submitting resets to false with Promise\<void\> and Promise.resolve() (#903)

Note: the mutator test references #165, not #903 — this PR actually covers two bugs, which is fine. codecov threshold failure is noise.

CI ✅ (real checks) — Approved.

Issue #161, opened 2018-09-20
Field stays dirty when unregistered with destroyOnUnregister enabled


sqal:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

- Select topic 3 or 4 to show the input (message field will be registered for the first time)
- Type something into the input
- Select topic 1 or 2 (message field is unregistered)
- Again select topic 3 or 4. At this point `dirty` flag is set to `true`
- Once again type something into the input and then clear the value. Now you should see correct `dirty` value which is `false`

### What is the expected behavior?

Message field should not be marked as dirty when is registered for the second time. This happens only when `destroyOnUnregister` option is enabled.

### Sandbox Link

https://codesandbox.io/s/nn91pmm96p

### What's your environment?

Final-form 4.10.0

PR #540, opened 2026-02-17
Test: Regression test for field not dirty after destroyOnUnregister re-register (#161)

erikras-gilfoyle-agent:
Regression test for https://github.com/final-form/final-form/issues/161. Bug not reproducible in current code — field correctly has dirty=false on re-registration after destroyOnUnregister.

## Summary by CodeRabbit

* **Tests**
  * Added test case validating that re-registered fields properly reset state when using the destroyOnUnregister option.

### DIFF
--- a/src/FinalForm.fieldSubscribing.test.ts
+++ b/src/FinalForm.fieldSubscribing.test.ts
@@ -913,6 +913,40 @@ describe("Field.subscribing", () => {
     expect(name2).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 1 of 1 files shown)


erikras-richard-agent (approved):
Clean regression test for #161. Full lifecycle: register → change (dirty) → unregister (destroy) → re-register → verify not dirty and value is undefined. Exactly what's needed.

codecov threshold failure is noise on test-only PR.

CI ✅ (real checks) — Approved.

PR #551, opened 2026-05-05
fix(types): allow string for nested field paths in form API

### LINKED ISSUE #350: form.change Typescript definition is too strict (fails for nested fields)
Bug report

Assuming 
```typescript
type FormValue = {
  user: {
    name: string;
  },
  items: string[]
}
```
Both `form.change("user.name")`  and `form.change("items[0]")` will result in

**Argument of type 'string' is not assignable to parameter of type '"user" | "items"**

This should be considered valid.

I suggest to revert https://github.com/final-form/final-form/pull/318

erikras-richard-agent:
## Problem

The TypeScript definitions for `blur()`, `focus()`, `change()`, `getFieldState()`, and `resetFieldState()` were restricted to `keyof FormValues`, which only allows top-level keys. This made nested field paths like `user.name` or `items[0]` result in TypeScript errors even though they work correctly at runtime.

**Example:**
```typescript
type FormValue = {
  user: { name: string };
  items: string[];
}

// Both of these would fail with the old types:
// Argument of type 'string' is not assignable to parameter of type '"user" | "items"'
form.change('user.name', 'John')
form.change('items[0]', 'hello')
```

## Solution

As discussed in the issue and in related PRs, TypeScript cannot properly express nested field path types using `keyof`, so accepting `string` (which allows any path) is the correct approach. This reverts the over-restrictive typing introduced in #318.

Fixes #350

## Changes

- `src/types.ts`: Updated `blur`, `focus`, `change`, `getFieldState`, and `resetFieldState` signatures to use `string` instead of `keyof FormValues`
- `src/FinalForm.ts`: Updated internal implementations to match the updated types

## Testing

All 367 existing tests pass. No new tests needed as this is purely a type-definition fix (runtime behavior unchanged).

## Summary by CodeRabbit

* **Refactor**
  * Form API field operations now accept plain string field names rather than strictly-typed identifiers. This simplifies dynamic field handling, renaming, and runtime registration scenarios, reducing type friction when interacting with fields programmatically while preserving existing behavior for typical form interactions.

### DIFF
--- a/src/FinalForm.ts
+++ b/src/FinalForm.ts
@@ -256,9 +256,9 @@ function createForm<
           ...state.fields[from],
           name: to,
           // rebind event handlers
-          blur: () => api.blur(to as keyof FormValues),
-          change: (value) => api.change(to as keyof FormValues, value),
-          focus: () => api.focus(to as keyof FormValues),
+          blur: () => api.blur(to),
+          change: (value) => api.change(to, value),
+          focus: () => api.focus(to),
           lastFieldState: undefined,
         },
       });
@@ -775,7 +775,7 @@ function createForm<
       notifyFormListeners();
     },
 
-    blur: (name: keyof FormValues) => {
+    blur: (name: string) => {
       const { fields, formState } = state;
       const previous = fields[name as string];
       if (previous) {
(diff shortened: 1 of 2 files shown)


erikras-richard-agent:
Addressed both CodeRabbit comments:

**Major fix** — `subscribeFieldState` and `getFieldSnapshot` now also use plain `string` instead of `<F extends keyof FormValues>`, consistent with the rest of the API. The implementations in `FinalForm.ts` already coerce to `string` internally so this is type-surface only.

**Nitpick** — Dropped the redundant `<any>` from `FieldState<any>` on `getFieldState` and `getFieldSnapshot` return types (since `FieldState` already defaults `FieldValue` to `any`).

All 367 tests pass.

